Issue #7, opened 2016-11-22
Evaluate SSDP/DIAL as a discovery mechanism


mfoltzgoogle:
See the meta-issue #5 for the list of evaluation criteria.

louaybassbouss:
FYI I just assigned this task to myself. I will put evaluation results in a markdown file in the home folder of this repository.

louaybassbouss:
PR #20 contains the initial version of SSDP evaluation. what still TODO is to have concrete numbers for Network and Power efficiency.

mfoltzgoogle:
Closing as this contains a good outline.  We can track some specific questions in followup issues. 

Filed #23 for getting power requirements.

PR #20, opened 2017-03-12
create a new document ssdp.md that evaluates SSDP protocol

louaybassbouss:
solves issue #7


ssdp.md
```diff
@@ -0,0 +1,135 @@
...
...g and Play) Device Architecture](http://www.upnp.org/specs/arch/UPnP-arch-DeviceArchitecture-v1.1.pdf). It allows control points (term used in UPnP architecture for control devices like smartphones or tablets) to search for devices or services of interest on the network. The SSDP protocol is also used in other standards/specifications without using the whole UPnP stack. One example is the [DIAL (Discovery and Launch)](http://www.dial-multiscreen.org/) which is a protocol that allows second-screen devices like smartphones or tablets to discover and launch applications on first-screen devices like TVs. Another example is [HbbTV 2.0 (Hybrid broadcast broadband TV)](https://www.hbbtv.org/wp-content/uploads/2015/07/HbbTV-SPEC20-00023-001-HbbTV_2.0.1_specification_for_publication_clean.pdf) which uses DIAL as underlying protocol to discover HbbTV devices and launch HbbTV applications. In [Physical Web Project](https://github.com/google/physical-web) there is a [proposal](https://github.com/google/physical-web/blob/master/documentation/ssdp_support.md) from [Fraunhofer FOKUS](https://www.fokus.fraunhofer.de/fame) on how to use the SSDP protocol to advertise and find URLs in local network.
```
mfoltzgoogle: Can you insert a couple of paragraph breaks to improve readability? Suggest:
- Before "One example"
- Before "Another example"
- Before "In Physical Web"

ssdp.md
```diff
@@ -0,0 +1,135 @@
...
+
+
+# Design and Specification
```
mfoltzgoogle: Maybe use a subheading (##) here as these following sections relate to SSDP as a whole?

ssdp.md
```diff
@@ -0,0 +1,135 @@
...
+
+SSDP allows root devices (term used in UPnP architecture for devices that offer services like TVs, printers, etc.) to advertise their services to control points on the network. It also allows control points to search for devices or services of interest at any time. SSDP specifies the discovery messages that are exchanged between control points and root devices. SSDP messages contain specific information about the service/device like type, unique device identifier, etc. SSDP uses part of the header field format of HTTP 1.1 ,but it is not based on full HTTP 1.1 as it uses UDP instead of TCP and it has its own processing rules. The following sequence diagram shows the SSDP message exchange between a control point and root device.
```
mfoltzgoogle: Suggest paragraph breaks before "SSDP messages" and "The following"

ssdp.md
```diff
@@ -0,0 +1,135 @@
...
+
+![](images/ssdp.png) 
+
+Message Flow:
+1. the root device advertises its services on the network by sending for each service it offers a "NOTIFY" message from type "ssdp:alive" to the multicast address "239.255.255.250:1900" with all the necessary information needed to access the service. Control points listening on the multicast address receive the message and check if the service is relevant for them or not. The NT header can be used to check the type of the service. To access more information about the service/device the control point needs to access the device description (format is specified in UPnP) by making an HTTP request to the URL provided in the LOCATION header of the SSDP message. The device description is a XML document that contains information about the device/service like friendly name, capabilities, etc.
```
mfoltzgoogle: It might be helpful to illustrate with NOTIFY, SEARCH and response messages for a hypothetical device.

ssdp.md
```diff
@@ -0,0 +1,135 @@
...
+1. a control point can search for root devices at any time by sending a SEARCH message to the multicast address. The search request contains the ST (search target) header specify the type of the service the control point is looking for. All devices listing to the multicast address will receive the search message.
+1. when a root device receives a search message and if it offers the requested service (by checking the ST header), it replies with a SEARCH response message only to the control point that sends the request (unicast).
+1. when a service is no more available, the root device needs to advertise a "NOTIFY" message from type "ssdp:byebye" with information about the corresponding service. Control points can remove the service if it is listed. 
+
+# Evaluation
+
+this section evaluates SSDP as discovery protocol for the OpenScreenProtocol according several functional and non-functional requirements
```
mfoltzgoogle: s/this/This/
s/OpenScreenProtocol/Open Screen protocol/

ssdp.md
```diff
@@ -0,0 +1,135 @@
...
+* Each display available on the network and runs a presentation receiver service replies to the search request with a SSDP message similar to the alive message.
+* The controller reads the LOCATION header for each available display and makes a HTTP GET request to get the device description XML. 
+* The controller parses each device description XML, reads the friendly Name of the display and checks if the display can open one of the URLs passed as input to the PresentationRequest constuctor. If yes, the device will be added to the list of available displays.
+* Open questions:
+    * how to send the endpoint of the receiver service: one solution is to use a HTTP header parameter in the response of the HTTP GET request for device description. DIAL uses this solution to send the endpoint of the DIAL server in the `Application-URL` HTTP response header. Another solution is to extend the XML device description with a new element to define the endpoint.
```
mfoltzgoogle: s/how/How/

ssdp.md
```diff
@@ -0,0 +1,135 @@
...
+
+* New display with presentation receiver service that appears on the network advertises the following SSDP alive message. The new header `FRIENDLY-NAME.example.org` is used for friendly name and `PRESENTATION-ENDPOINT.example.org` for endpoint of the receiver service. The value `urn:example-org:service:osp:1` of header `NT` is used for presentation receiver service.
```
mfoltzgoogle: Is `example.org` associated with the organization standardizing the names of these OSP-specific headers?  Or is it associated with the manufacturer of the display?

ssdp.md
```diff
@@ -0,0 +1,135 @@
...
+
+* A controller sends the following SSDP search message to the multicast address. The new header `PRESENTATION-URLS.example.org` allows the controller to send the presentation URLs to the display. 
+
+    ```
+    M-SEARCH * HTTP/1.1
+    HOST: 239.255.255.250:1900
+    MAN: "ssdp:discover"
+    MX: seconds to delay response
+    ST: urn:example-org:service:osp:1
+    PRESENTATION-URLS.example.org: http://example.com/foo.html, http://example.com/bar.html
```
mfoltzgoogle: Including presentation URLs in the SEARCH or response has a couple of issues:

- It's not great from a privacy perspective, as it exposes them to passive network listeners.
- These messages need to fit in a UDP packet, which limits them to about 1420 bytes.

ssdp.md
```diff
@@ -0,0 +1,135 @@
...
+    ```
+    NOTIFY * HTTP/1.1
+    HOST: 239.255.255.250:1900
+    CACHE-CONTROL: max-age = seconds until advertisement expires
+    LOCATION: URL of the device description
+    NTS: ssdp:alive
+    SERVER: OS/version UPnP/1.0 product/version
+    USN: advertisement UUID
+    NT: urn:example-org:service:osp:1
+    FRIENDLY-NAME.example.org: My Presentation Display
```
mfoltzgoogle: How is FRIENDLY-NAME encoded?

ssdp.md
```diff
@@ -0,0 +1,135 @@
...
+## Non-functional Requirements
+
+### Reliability
+
+UPnP recommend due to the unreliable nature of UDP to send the SSDP messages more than once (and max three times to avoid network congestion) with some delay of few hundred milliseconds. In addition, the device must re-send its advertisements periodically prior to expiration of the duration specified in
+the `CACHE-CONTROL` header (minimum value is 1800s).  
+
+### Latency of device discovery / device removal
+
+New devices added or removed can be immediately detected if the controller listens to the multicast address for "ssdp:alive" and "ssdp:byebye" messages. For search requests the latency depends on the `MX` SSDP header which contains the maximum wait time in seconds. According to UPnP specification, it must be greater than or equal to 1 and should be less than 5 inclusive. SSDP responses should be delayed a random duration between 0 and the value of `MX` to balance load for the controller when it processes responses.
```
mfoltzgoogle: Do you have a sense of how long removal detection takes in the easy case (TV shut off and sends byebye) and the hard case (shut off without any notification?)

ssdp.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Security - both of the implementation, and whether it can be leveraged to enhance security of the entire protocol
+
+SSDP consider local network as secure environment. Any device on the network can discover other devices or services. Security can be implemented on service level. In our case the output of the discovery is a list of displays that contain friendly names and endpoints of presentation receiver services. Any security mechanism can be implemented during session establishment and communication independent from the method used for discovery. 
+
+### Privacy: what device information is exposed
+
+The standard UPnP device description exposes parameters about the device/service like unique identifier, friendly name, software, manufacturer, service endpoints, etc.
```
mfoltzgoogle: We should think about whether it's necessary to expose the friendly name (or even the UUID) in the network advertisements.  If the controlling user agent just needs a list of IP:port of potential receiver services, maybe they can be dropped.

ssdp.md
```diff
@@ -0,0 +1,135 @@
...
+    ```
+
+### Remote Playback API: Monitor Display Availability
+TODO (it should be very similar to the Presentation API display availability but for capabilities other parameters related media formats, codecs, encryptions need to be considered)
+
+## Non-functional Requirements
+
+### Reliability
+
+UPnP recommend due to the unreliable nature of UDP to send the SSDP messages more than once (and max three times to avoid network congestion) with some delay of few hundred milliseconds. In addition, the device must re-send its advertisements periodically prior to expiration of the duration specified in
```
mfoltzgoogle: Chrome desktop sends three search requests spaced 2 seconds apart every 120 seconds.  We've found this is good enough from a reliability perspective, and hope to share some more concrete data.

ssdp.md
```diff
@@ -0,0 +1,135 @@
...
+
+* The display advertises the presentation receiver service when it is added on the network using SSDP and the service type `urn:example-org:service:osp:1` (this is just an example). The SSDP alive message contains a LOCATION header which points to the XML device description (contains friendly name, device capabilities, etc.). The display should advertise a SSDP byebye message before it is not more available. 
+* The controller starts "Monitor Display Availability" algorithm by sending a SSDP search message with the service type `urn:example-org:service:osp:1` and waits for responses from devices running a presentation receiver service. The controller should wait for SSDP alive/byebye messages on the multicast address to keep the list of available displays up-to-date (e.g. when a new display in added or an existing one is removed).  
+* Each display available on the network and runs a presentation receiver service replies to the search request with a SSDP message similar to the alive message.
+* The controller reads the LOCATION header for each available display and makes a HTTP GET request to get the device description XML.
```
mfoltzgoogle: Adding a dependency on HTTP+XML adds some security risks and complexity from an implementer perspective.

- Untrusted XML parsing is generally unsafe and should be done in a sandboxed process.
- The HTTP requests have to be done carefully since they bypass all of the normal origin related checks that you would get from Web content.

ssdp.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Network efficiency
+
+It depends on multiple factors like the number of devices in the network using SSDP (includes devices that support DLNA, DIAL, HbbTV 2.0, etc.), the number of services provided by each device, the interval to re-send refreshment messages (value of `CACHE-CONTROL` header), the number of devices/applications sending discovery messages.
+
+### Power efficiency
+
+This depends on many factors like the used method (see section [Presentation API: Monitor Display Availability](#presentation-api-monitor-display-availability)), number of devices in the Network using UPnP/DIAL and the way how to monitor the display availability.
+
+* regarding the used the method, "Method 2" is better than "Method 1" regarding power efficiency. In Method 1 the controller needs to create and send SSDP search requests, receive and parse SSDP messages, make HTTP requests to get device descriptions and parse device description XML to get friendly name and check capabilities. In method 2 the controller needs only to create and send search requests and receive and parse SSDP messages.
```
mfoltzgoogle: Regarding the method used,

ssdp.md
louaybassbouss: Completely agree this is why I proposed in Method2 a more lightweight solution without using HTTP and XML. But Method1 illustrates the usage of SSDP in UPnP, DIAL, etc.

ssdp.md
louaybassbouss: It depends from the implementation of SSDP on the controller. For the case "TV shut off and sends byebye" and the controller is listening on the multicast address "239.255.255.250:1900" the device removal will be detected as soon as the byebye message is received and parsed (It will take maybe few milliseconds).  For the case "shut off without any notification" and the controller is listening on the "239.255.255.250:1900" the controller can only detect the device removal after the last advertisement received from that device expires (`CACHE-CONTROL` header contains the expiration in seconds).  

 If the controller is not listening on the multicast address "239.255.255.250:1900" there is no difference between the two cases you mentioned and the device removal can only be detected after the controller sends the next SEARCH message and compare the response messages with the old list of available devices.

ssdp.md
louaybassbouss: I think it is very important to clearly define what is the input and output of the discovery process. Exposing friendly name reduce the number of calls needed to show the Display Selection Dialog on the client. UUID helps in case of SSDP to detect device removal or detect messages from same device.

ssdp.md
louaybassbouss: great we can update this section later if you can share more concrete data

ssdp.md
louaybassbouss: I agree with you question is as I mentioned in other comments we need to make clear what is the input and output of the discovery process. Question is here do we want to have device filtering in discovery or later after handshake? filter devices during discovery reduce the number of unnecessary calls (to displays that cannot present a specific URL) but raises on the other hand couple of privacy issues as you mentioned. From my point of view, I prefer a solution where the presentation display exposes in a way more information about the content that can be presented. I was thinking about a regular expression that can be exposed in a SSDP header (in SEARCH response or alive messages) where the controller can detect if the display can present a specific URL or not. For example if a display allows only https then the regular expression will be something like `https://*` or if the receiver is cast device a accepts only cast applications then the regular expression could be `https://cast.google.com#castAppId=*`. In this case there is no need to share the presentation URLs in the discovery messages.

ssdp.md
louaybassbouss: with the organization standardizing the names of the OSP-specific headers.

louaybassbouss:
thx @mfoltzgoogle for the feedback I will submit a new PR addressing your comments.

ssdp.md
louaybassbouss: UPDATE: DIAL introduces in version 2.0 a new SSDP header `WAKEUP` in SEARCH responses from presentation devices that support WoL. @mwatson2 do yo have more details on this? is `WAKEUP` introduced in DIAL oder somewhere else?

ssdp.md
louaybassbouss: it is just a clear-text string field like the other SSDP headers.

ssdp.md
louaybassbouss: done

louaybassbouss:
@mfoltzgoogle the last commit addresses your comments.

ssdp.md
mfoltzgoogle: Since it's displayed to humans, it should be able to represent Unicode code points.  We can look at the SSDP spec to see what is supported by default.

ssdp.md
mfoltzgoogle: I think if there's a privacy preserving way to optimize away some connections to the receiver service that would be a benefit.   If the primary benefit is to separate out generic endpoints (https://*)  from application-specific ones (e.g., casts://*) a simple scheme advertisement approach might work.

mfoltzgoogle: Filed https://github.com/webscreens/openscreenprotocol/issues/21

ssdp.md
mfoltzgoogle: Filed https://github.com/webscreens/openscreenprotocol/issues/22

mfoltzgoogle:
Thanks again @louaybassbouss.  There are a few small editorial items I can address in a followup PR.  I filed issues for a couple of the threads that came up when discussing the content as well.

Issue #5, opened 2016-11-22
[Meta] Evaluate presentation screen discovery mechanisms


mfoltzgoogle:
This issue covers evaluation of presentation screen discovery mechanisms.

Per the charter scope, the focus will be on LAN discovery mechanisms.  The following have been proposed thus far:

- [x] Multicast DNS/DNS-SD #6
- [x] SSDP/DIAL #7

(If there are any I have missed, please reply to this issue.)

The task for each is to evaluate them along the functional and non-functional requirements, and make an initial recommendation to explore further.

mfoltzgoogle:
The evaluations should cover the following criteria (non-exhaustive):

- Functional
- Reliability
- Ease of implementation / deployment
- Security - both of the implementation, and whether it can be leveraged to enhance security of the entire protocol
- Privacy: what device information is exposed 
- Network efficiency
- Power efficiency
- IPv4 and IPv6 support
- Standardization status

louaybassbouss:
@mfoltzgoogle you can count on me for the discovery part. I will first start with SSDP/DIAL.

mwatson2:
@louaybassbouss Please let me know if I can help with the DIAL related aspects.

louaybassbouss:
@mwatson2 sorry for late reply, yes I will let you know if I need help. I am not sure if in the evaluation of discovery protocols we need to consider DIAL or only SSDP. I think the DIAL API part (POST/GET/DELETE HTTP requests on the /apps/... endpoint) could be evaluated in the "Protocol Signalling". What do you think?

chrisn:
As input to the evaluation, here's some [previous work](https://github.com/bbc/device-discovery-pairing/blob/master/document.md) from myself and a colleague - there may be some useful info on mDNS and SSDP.

louaybassbouss:
@mfoltzgoogle @mwatson2 @chrisn I just submitted the PR #20 which contains the SSDP evaluation. feel free to review and comment.
@mwatson2 I focussed in the document only on the discovery part which is SSDP but discussed how DIAL uses SSDP, Device Description and how it signal the Endpoint of the DIAL Server. We should consider DIAL REST API in the control protocol (#11) evaluation as well. What do you think?

mfoltzgoogle:
We've landed initial evaluations for #6 and #7.  There are some follow up items around security and data collection which are tracked under separate issues linked from the main documents.

[...]

Issue #6, opened 2016-11-22
Evaluate Multicast DNS/DNS-SD as a discovery mechanism


mfoltzgoogle:
See the meta-issue #5 for the list of evaluation criteria.

@mfoltzgoogle I will take this work item to prepare a report .

mfoltzgoogle:
Went ahead and landed a report.  Feel free to open issues or send email with followup comments/questions.

PR #35, opened 2017-05-26
Adds template for discovery & tweak transport template

mfoltzgoogle:
This adds a template for evaluation of network discovery protocols (Issue #5) and tweaks the transport template to be consistent with that.


mfoltzgoogle:
Merging so I can adopt this to work on Issue #6: Evaluate Multicast DNS/DNS-SD as a discovery mechanism

PR #36, opened 2017-05-30
Reformat ssdp.md according to templates/discovery.md

mfoltzgoogle:
FYI @louaybassbouss 

This PR reformats and reorders the sections of ssdp.md according to the template committed in PR #35.

It also adds a short note about hardware requirements which was missing in the original document.


louaybassbouss:
thx @mfoltzgoogle LGTM I will merge this PR

Issue #21, opened 2017-03-27
[SSDP] Investigate mechanisms to pre-filter devices by Presentation URL


mfoltzgoogle:
Forked from https://github.com/webscreens/openscreenprotocol/pull/20/files#r108024628.

---

Including presentation URLs in the SEARCH or response has a couple of issues:

* It's not great from a privacy perspective, as it exposes them to passive network listeners.
* These messages need to fit in a UDP packet, which limits them to about 1420 bytes.

@louaybassbouss

louaybassbouss 2 days ago Collaborator

I agree with you question is as I mentioned in other comments we need to make clear what is the input and output of the discovery process. Question is here do we want to have device filtering in discovery or later after handshake? filter devices during discovery reduce the number of unnecessary calls (to displays that cannot present a specific URL) but raises on the other hand couple of privacy issues as you mentioned. From my point of view, I prefer a solution where the presentation display exposes in a way more information about the content that can be presented. I was thinking about a regular expression that can be exposed in a SSDP header (in SEARCH response or alive messages) where the controller can detect if the display can present a specific URL or not. For example if a display allows only https then the regular expression will be something like https://* or if the receiver is cast device a accepts only cast applications then the regular expression could be https://cast.google.com#castAppId=*. In this case there is no need to share the presentation URLs in the discovery messages.

@mfoltzgoogle
mfoltzgoogle just now Collaborator
I think if there's a privacy preserving way to optimize away some connections to the receiver service that would be a benefit. If the primary benefit is to separate out generic endpoints (https://) from application-specific ones (e.g., casts://) a simple scheme advertisement approach might work.

mfoltzgoogle:
Assuming that PR #59 is considered a good approach, we should apply it to mDNS as well (via a TXT record).

mfoltzgoogle:
Closing as:

- We are not using SSDP any more so discussion of SSDP is moot.
- There is a standard way to advertise device-specific capabilites through protocol extensions: https://w3c.github.io/openscreenprotocol/#protocol-extensions

Issue #10, opened 2016-11-22
Evaluate Mozilla proposal as a protocol transport


mfoltzgoogle:
This work item tracks the evaluation of the Mozilla proposal [1] as a transport for the Open Screen protocol.

See the meta issue #8 for the list of criteria.

[1] https://wiki.mozilla.org/WebAPI/PresentationAPI:Protocol_Draft#Establish_Control_Channel

@schien Can you take stab at this when you are available?

mfoltzgoogle:
I think this can be closed.  Followup items to add to the evaluation are tracked under separate issues.

Issue #9, opened 2016-11-22
Evaluate QUIC as a protocol transport


mfoltzgoogle:
This issue tracks the work to evaluate QUIC [1] as an alternative transport for the Open Screen protocol.

See the meta issue #8 for the list of evaluation criteria.

[1] https://www.chromium.org/quic

mfoltzgoogle:
I'll take on this work item.

mfoltzgoogle:
I've merged in a draft evaluation using the template in https://github.com/webscreens/openscreenprotocol/pull/16.  It needs refinement, which will happen once we make progress on the API level protocols.  Also need to work on getting some hard data around latency and performance.

mfoltzgoogle:
Meant to add: we can track additional work items on QUIC in separate issues as needed.  I'll file some placeholders now.

Closing this for now.

PR #44, opened 2017-08-29
Issue 11 control protocol

mfoltzgoogle:
PTAL @schien @anssiko @tidoust @chrisn 

This is a draft of a wire format for the control protocol for the Presentation API.  It defines a generic message format supporting commands, events, and request/response pairs.  It should be possible to layer it on top of QUIC or an RTCDataChannel (with a few modifications in either case).

I define a few example messages using this format for presentation screen availability and presentation initiation.  I'll take any feedback on this draft and fill out the rest of the message definitions in another pull request.

If the feedback asks for major changes, then I will recommend that the author submit a separate side-by-side proposal :-)

### DIFF
--- a/control_protocol.md
+++ b/control_protocol.md
@@ -1,12 +1,19 @@
-# Open Screen Control Protocols
+# Open Screen Control Protocol
 
-This document outlines a control protocol for the Presentation API and the
-Remote Playback API.  It is intended to be implemented on top of one of the
-proposed transport mechanisms such as [QUIC](quic.md)
-or [RTCDataChannel](datachannel.md).  The control protocol is responsible for
-mapping Web API operations (including application messages) onto a network
-syntax for transmission between controlling user agents and presentation
-screens, and ultimately between presentation controller and receiver pages.
+This document outlines a control protocol suitable to support the implementation
+of the [Presentation API](https://w3c.github.io/presentation-api/)
+and the [Remote Playback API](https://w3c.github.io/remote-playback/).
+It is intended to be implemented on top of one of the proposed Open Screen
+transport mechanisms such as [QUIC](quic.md) or [RTCDataChannel](datachannel.md).
+
+The control protocol is responsible for mapping Web API operations and data
+provided by Web applications onto network messages for transmission between
+controlling user agents and receiving user agents (or remote playback devices).
+
+This document focuses on the syntax and semantics of the network messages
+themselves.  The exact content and sequencing of the messages in response to Web
+API behavior (to implement the steps of the Presentation API and Remote Playback
+API) will be explained in future updates to this document.
 
 ## Requirements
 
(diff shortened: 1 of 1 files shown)


tidoust:
This looks very good and well documented!

One thing that we already discussed in the past (although I'm not clear we came to a conclusion) is that, if I understand correctly, the suggested "Presentation Display Availability Request" will leak presentation request URLs to discovered receivers at a stage where the user has not yet given her consent to use any specific receiver, that is before the user selects the receiver.

Now this approach has the advantage of allowing any kind of suport logic in receivers. Any alternative would be much less flexible, as it would require a declarative mechanism that receivers would use to explain what types of URLs they support, and which the controller would in turn use to decide whether a given receiver is likely compatible with a given URL.

PS: FWIW, there are a few typos here and there. I'll try to point them out when I have more time available.

anssiko:
Great draft! You may complete the rest of the message definitions.

How confident we are at this point after completing the evaluation of QUIC #9 and RTCDataChannel #10 (aka Mozilla's proposal) that these two transports are the ones this control protocol is to be layered on top off, and not others? I've understood for both of these transports, there's implementers' interest, which is a good sign.

Is anyone thinking of using HTTP/2, WebSockets, or a custom protocol anymore? The sooner we're able to nail down the set of transports to target, the easier it will be to make progress on the control protocol. So I'm wondering if we could nail down the supported transports now to ensure the constraints in protocol design would be known upfront.

control_protocol.md
```diff
@@ -0,0 +1,478 @@
...
+
+## Message Transport
+
+To meet the requirements above, a message oriented control protocol is
+necessary.  That means that each party should be able to transmit a sequence of
+variable-length messages from the other, and have every message received by the
+other party, intact and in-order.
+
+The RTCDataChannel does support variable length messages (**TODO:** find a spec
+for it!), although there are
```
tomoyukilabs: If I understand correctly, [draft-ietf-rtcweb-data-channel-13](https://tools.ietf.org/html/draft-ietf-rtcweb-data-channel-13) defines the message limits of data channels. [Section 4](https://tools.ietf.org/html/draft-ietf-rtcweb-data-channel-13#section-4) defines the requirements:

> Req. 8:   The data channel transport protocol should support unbounded-length "messages" (i.e., a virtual socket stream) at the application layer, for such things as image-file-transfer; Implementations might enforce a reasonable message size limit.

On the other hand, [Section 6.6](https://tools.ietf.org/html/draft-ietf-rtcweb-data-channel-13#section-6.6) illustrates message interleaving to avoid monopolization of data channel:

> The SCTP base protocol specified in [RFC4960] does not support the interleaving of user messages.  Therefore sending a large user message can monopolize the SCTP association.  To overcome this limitation, [I-D.ietf-tsvwg-sctp-ndata] defines an extension to support message interleaving, which SHOULD be used.  As long as message interleaving is not supported, the sender SHOULD limit the maximum message size to 16 KB to avoid monopolization.

control_protocol.md
```diff
@@ -0,0 +1,478 @@
...
+  identify messages originating from one party, and to ensure that messages are
+  handled in the proper order by the recipient.
+- `REQUEST_ID`: An optional 64-bit unsigned integer.  If the message flavor is a
+  Response, it contains the `SEQUENCE_ID` of the Request that the response is
+  replying to.  For other flavors, it is not present.
+
+The content of the `MESSAGE_BODY` is not constrained by the generic message
+structure, and must be interpreted according to the `MESSAGE_TYPE`.
+
+**TODO:** Investigate variable length integers for MESSAGE_LENGTH, SEQUENCE_ID
```
chrisn: A starting point for the investigation: https://en.wikipedia.org/wiki/Variable-length_quantity. Includes links to particular encodings, e.g, Protocol Buffers. Should we raise as a separate issue for discussion?

tomoyukilabs:
The draft looks great!

A minor comment: One of my interests is how to pass preferred locale through the control protocol. I expect that such kind of parameters could optionally be attached to appropriate request messages (e.g. Presentation Initiation Request). Is it correct?

chrisn:
This is great, thank you Mark!

@anssiko said:

> Is anyone thinking of using HTTP/2, WebSockets, or a custom protocol anymore? 

This is something we should discuss with the Media and Entertainment IG, in relation to TV industry second screen protocols. For example, the companion screen protocol in HbbTV 2.0 uses WebSockets (with SSDP and DIAL) but also UDP for messages involved in synchronising content between screens.

mfoltzgoogle:
@tidoust Yes, the ability to "pre-filter" receivers by presentation URL is not included in this first draft.  https://github.com/webscreens/openscreenprotocol/issues/21 suggests that this could be done in discovery, or later after a control protocol is established, and the exact filtering mechanism isn't decided yet.   I will add a TODO referencing that issue near the presentation availability request definition.

mfoltzgoogle:
@anssiko @chrisn No evaluation has yet taken place of HTTP/2 or WebSockets as potential transports. From my point of view, I won't be proposing them unless they provide a capability missing from QUIC or RTCDataChannel that can't be addressed at the control protocol level.

It would be helpful to reach out to the M&E IG on this.  If they are invested in these mechanisms and want to pursue them for open screen, we can attempt a deeper dive.

control_protocol.md
mfoltzgoogle: Good idea. Filed https://github.com/webscreens/openscreenprotocol/issues/45 and will link from this TODO.

mfoltzgoogle:
@tomoyukilabs Good point - this was missed in the initial definition of the presentation initiation request.  Will update in the next commit.

control_protocol.md
mfoltzgoogle: Thanks.  I think I was looking for [SCTP](https://tools.ietf.org/html/rfc4960) as the layer that provides message framing.

anssiko:
>It would be helpful to reach out to the M&E IG on this. If they are invested in these mechanisms and want to pursue them for open screen, we can attempt a deeper dive.

I tried to move this discussion to the mailing list https://lists.w3.org/Archives/Public/public-secondscreen/2017Aug/0063.html and asked if @chrisn could bring this topic to the M&E IG.

Agreed that the first good step for M&E IG participants interested in other transports would be to do similar evaluation that has been done already for QUIC #9 and RTCDataChannel #10.

control_protocol.md
```diff
@@ -0,0 +1,500 @@
...
+adopting [SCTP](https://tools.ietf.org/html/rfc4960) as its message transmission
+protocol. The
+[RTCDataChannel specification](https://tools.ietf.org/html/draft-ietf-rtcweb-data-channel-13)
+suggests there may be
+[implementation restrictions on message size](https://tools.ietf.org/html/draft-ietf-rtcweb-data-channel-13#section-4)
+and that implementations should
+[interleave messages](https://tools.ietf.org/html/draft-ietf-rtcweb-data-channel-13#section-6.6)
+for fairness.
+
+Meanwhile, QUIC is stream-oriented and not message oriented, so a message
```
schien: Since QUIC is stream-oriented, it'll have hand-of-line blocking issue while transmitting large message. Maybe we can introduce additional out-of-band channel, this can be discussed in another issue.

control_protocol.md
```diff
@@ -0,0 +1,500 @@
...
+               .                       .
+               +-----------------------+
+               + AVAILABILITY_RESULT_N +
+               +-----------------------+
+```
+
+- `NUM_URLS` is an unsigned positive 16-bit integer that is identical to
+  `NUM_URLS` in the corresponding Presentation Display Availability Request.
+- Each `AVAILABILITY_RESULT` is an unsigned 8-bit integer containing the
+  availability result for the Nth URL in the request.  The results are as follows:
```
schien: Receiver side might have its own preference of the compatible URLs. We can extend this message by sending a list of (index, result) tuples.

control_protocol.md
```diff
@@ -0,0 +1,500 @@
...
+  that communicates the ID for the presentation.  Values shorter than 128 bytes
+  should be zero-byte padded.
+- `ACCEPT_LANGUAGE` is a null (zero-byte) terminated ASCII string of 128 bytes
+  that communicates the preferred locale for the presentation.  It is in the
+  format defined by the `Accept-Language:` header for HTTP, defined
+  in [RFC 7231](https://tools.ietf.org/html/rfc7231#section-5.3.5).
+  Values shorter than 128 bytes should be zero-byte padded.
+- `URL_LENGTH` is an unsigned positive 32-bit integer with the length, in bytes,
+  of the presentation URL.
+- `URL_CONTENT` is the presentation URL, encoded according to RFC 3986.
```
schien: Do we want to allow arbitrary HTTP header to be override, i.e. make receiver send particular cookie as sender to simplify the authentication procedure?

control_protocol.md
mfoltzgoogle: QUIC is designed to prevent this from happening by multiplexing among streams between two endpoints.  The proposal submitted here suggests assigning one stream to each PresentationConnection, so a message sent on one connection shouldn't block a message sent on another.  This is something we can collect data on as well.

control_protocol.md
mfoltzgoogle: Good point, however, the current algorithm for starting presentation allows the controller to determine the URL priority:

https://w3c.github.io/presentation-api/#dfn-start-a-presentation-connection, step 5.

Meaning, if the page creates PresentationReuqest([URL1, URL2]), and the receiver reports [URL2, URL1] as compatible, the page must still initiate presentation with URL1.

However, I could see an argument for allowing the receiver to influence this priority, for example preferring pre-installed apps for some URLs.  So I'll add this enhancement in case we update the controller algorithms get updated.

control_protocol.md
mfoltzgoogle: That seems like a reasonable enhancement, although the API spec doesn't call for it now.  Done.

mfoltzgoogle:
Thanks for the feedback @schien.  I uploaded a commit with a couple of your suggested changes, and will now merge as I think everyone has provided feedback.

This is a working document so everyone should feel free to make further comments/suggestions on GH.

Issue #46, opened 2017-09-11
Editorial TODO: Add glossary reference and make language consistent


mfoltzgoogle:
When working on spec documents I've been using the terms controller, controlling page, receiver, presentation display, presentation, etc. interchangeably.

This TODO is a reminder to settle on a set of terms (from the requirements document), edit the other documents to be consistent with those terms, and possibly add a short glossary for reference.

PR #48, opened 2017-09-15
Define and use consistent terminology.

mfoltzgoogle:
Addresses Issue #46: Editorial TODO: Add glossary reference and make language consistent.

In requirements.md, the following terms are defined in a manner consistent with the Presentation API spec:
- controlling user agent
- controller
- receiver
- presentation
- presentation display

Relationships of these terms to parts of the protocol:
- discovery is a protocol between the _controlling user agent_ and the _presentation display_ to enable connection to a _receiver_
- transport is a connection between the _controlling user agent_ and the _receiver_
- the control protocol is between the _controlling user agent_ and the _receiver_
- presentation messages are between a _controller_ and _presentation_

This large but mechanical change applies this terminology throughout.  (It would be nice to have an architectural diagram summarizing this...)


mfoltzgoogle:
Going to merge this; if we change our minds on terminology, it should be simple to search and replace, now that terms are used consistently.

Issue #11, opened 2016-11-22
Presentation API control protocol


mfoltzgoogle:
This work item covers the control protocol for the Presentation API, i.e. the syntax and meaning of the messages sent over the transport [1] to start and control the presentation.

It might make sense to pick an initial transport before undertaking this work item, since the format of the control messages may be tied to the transport.  For example, we may be able to map control messages onto transport level mechanisms like QUIC resource streams.

[1] #8 Evaluate protocol transport mechanisms

mfoltzgoogle:
I'm closing this issue since we have a working definition drafted by the three pull requests for the Presentation API.   There are several TODOs in the document tracked under other GitHub issues.  I'll file one issue more to link specific messages to steps in the Presentation API spec so it's clearer how the semantics of the API map onto the control protocol.

PR #53, opened 2017-09-20
Adds Presentation Receiver status.

mfoltzgoogle:
Addresses Issue #11: Presentation API control protocol

Defines Presentation Reciever Status, the last message in the Presentation API control protocol.

Application Message:
- Adds a Locale field.
- Adds special handling for empty text messages.
- Defines encoding of text messages.
- Other corrections.

### DIFF
--- a/control_protocol.md
+++ b/control_protocol.md
@@ -675,10 +675,8 @@ Byte Offset
   164          +-----------------------+
                +  MESSAGE_TYPE         +
   165          +-----------------------+
-               +  MESSAGE_LOCALE       +
-  199          +-----------------------+
                +  MESSAGE_LENGTH       +
-  203          +-----------------------+
+  197          +-----------------------+
                +  MESSAGE_CONTENT      +
                +-----------------------+
 ```
@@ -696,10 +694,6 @@ Message Type | Meaning
 2            | Binary message
 3            | Empty text message
 
-- `MESSAGE_LOCALE` is an optional [BCP-47 language
-  code](https://tools.ietf.org/html/bcp47) that identifies the locale of a text
-  message.  It is a zero-terminated 64-byte ASCII string, zero padded on the
-  right.  If the `MESSAGE_TYPE` is not 1, this is all zeroes.
 - `MESSAGE_LENGTH` is an unsigned 32-bit integer with the length, in bytes,
   of the message content.
 - `MESSAGE_CONTENT` is the content of the message.  It must be exactly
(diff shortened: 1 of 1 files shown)


tomoyukilabs:
Regarding a locale, it looks almost good to me! Locales of display's friendly name and presentation title are very helpful to display them properly.

On the other hand, I wonder if the `MESSAGE_LOCALE` field in Presentation Application Message would be useful or not. Unless any method or parameter to specify a locale of each message is not be exposed to web applications, the field will eventually be set to the same string as specified in Presentation Initiation Request and will not vary until the presentation would be terminated.

By the way, I suppose that each locale field could have variable length. In most cases, a locale string is very short like `en-US`, `zh-CN`, `zh-TW`, `ko`, `ja`, etc.

mfoltzgoogle:
> On the other hand, I wonder if the MESSAGE_LOCALE field in Presentation Application Message would be useful or not. Unless any method or parameter to specify a locale of each message is not exposed to web applications, the field will eventually be set to the same string as specified in Presentation Initiation Request and will not vary until the presentation would be terminated

Good point.  I remember we discussed passing locale information with application messages, and deciding on leaving it up to the application, as it knows best which messages need a locale and what that locale is.  In the future, we could add an API like `send(<body>, <locale>)` to allow applications to pass a locale with a string to capture this pattern.

For now, I'll remove `MESSAGE_LOCALE` since I don't think either end has a good way to fill it.

> By the way, I suppose that each locale field could have variable length. In most cases, a locale string is very short like en-US, zh-CN, zh-TW, ko, ja, etc.

The BCP-47 spec doesn't seem to limit the length because of arbitrary variant strings that can be appended to the two letter codes.  64 characters captured all examples from the spec, but I didn't research this deeply.  We could restrict to a subset of BCP-47 if we want to ensure shorter codes.

Maybe the i18n WG would have some advice for us on this.

tomoyukilabs:
> In the future, we could add an API like send(<body>, <locale>) to allow applications to pass a locale with a string to capture this pattern.

I agree with that idea.

> 64 characters captured all examples from the spec, but I didn't research this deeply.

Okay. I just wanted to confirm the reason of _64_ characters. Regarding the length, BCP-47 describes [Length Considerations](https://tools.ietf.org/html/bcp47#section-4.4) and indicates several examples and recommendations, e.g.:

- historically most language tags have consisted of language and region subtags with a combined total length of up to six characters
- RFC4646 recommended a minimum field size of 42 characters

> Maybe the i18n WG would have some advice for us on this.

Absolutely.

mfoltzgoogle:
@tomoyukilabs 

> Okay. I just wanted to confirm the reason of 64 characters.

It was a best guess, and turns out to be consistent with the length guidelines in RFC4646.

In general, I'm using fixed length fields to constrain the message format when possible to simplify the code that constructs and parses these messages.  If there's a need for language tags longer than 64 characters we can convert it to a variable length string (with a 16-bit length field).

I'll add a TODO with a followup issue to research this more and/or request input from the i18n group when ready.

Issue #41, opened 2017-06-15
[Transport] Add section to evaluation indicating compatibility with guest mode


mfoltzgoogle:
Although specific mechanisms for WAN use are out of scope, it is useful to know if a transport can be extended in the future to support displays outside the controller's LAN, or is inherently tied to a specific LAN.

mfoltzgoogle:
Based on the thread in PR #56, we'll use this issue to cover a more general definition of "guest mode."  Since that's out of scope for v1, marking as a v2 work item.

PR #56, opened 2017-10-05
Add material around guest mode.

mfoltzgoogle:
Addresses Issue #41: [Transport] Add section to evaluation indicating compatibility with guest mode

Adds a section to the Transport template for guest mode, and discussion in two existing transports how they could implement guest mode.

### DIFF
--- a/datachannel.md
+++ b/datachannel.md
@@ -146,6 +146,26 @@ control message and data will be encrypted.
 
 Data Channel depends on SCTP over UDP, which is supported for both IPv4 and IPv6.
 
+# Guest mode support
+
+As part of WebRTC, Data Channel works in conjunction with the network traversal
+services in WebRTC ([STUN](https://tools.ietf.org/html/rfc5389),
+[TURN](https://tools.ietf.org/rfc/rfc5766) and
+[ICE](https://tools.ietf.org/html/rfc5245)) to enable connections between
+endpoints on different LANs.
+
+However, in order to use these services, the two parties must be able to
+exchange offer and answer messages as well as ICE candidates.  WebRTC does not
+define how this is done, but often a cloud service is used to facilitate this
+exchange.
+
+In addition, to operate STUN and TURN additional cloud services (STUN and TURN
+servers) are required.
+
+While feasible, additional APIs and specifications would be required to define
+how the controlling user agent and presentation display provision and gain
+access to the necessary cloud services.
+
 # Hardware requirements
 
 Major desktop, laptop, and high-end mobile phone should be able to create single
--- a/quic.md
+++ b/quic.md
@@ -130,6 +130,30 @@ information.
 
 QUIC depends only on UDP, which is supported for both IPv4 and IPv6.
 
+# Guest Mode support
+
+QUIC requires the ability route IP packets between both endpoints.  If both
+IP:port combinations are already exposed on the WAN, then no additional changes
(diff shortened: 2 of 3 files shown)


mfoltzgoogle:
PTAL @schien for Data Channel changes
PTAL @louaybassbouss who asked for guest mode evaluation

schien:
Overall this looks good to me. However I don't understand the relationship between WAN support and Guest mode. Is Guest mode not able to be used in LAN? Would be helpful if @mfoltzgoogle can provide some more details.

chrisn:
I agree with @schien's comment. I'm not sure what we mean by Guest mode - e.g., the [documentation for Google Cast](https://developers.google.com/cast/docs/guest_mode) mentions using BLE or ultrasound to communicate locally.

mfoltzgoogle:
@louaybassbouss made the original request to evaluate transports on "guest mode" feasibility and may be able to shed some light.

For this PR, I interpreted "guest mode" to mean: "How to make this transport work for two endpoints that are connected to two distinct LANs and not directly routable."  For example, a private network and a guest network, or a home WiFi network and a cellular data provider.

In general this requires routing the control channel through a WAN that interconnects the two LANs.  In almost all cases of interest "WAN" = "the Internet" and I can simplify the text accordingly.

There are a couple of other potential scenarios that one could call "guest mode":
- How devices that are not on the same LAN (or multicast-restricted because of client isolation) discover each other.  That can be done in a variety of ways including Bluetooth, audio beacons, cloud services etc. The Google Cast guest mode uses some of these techniques.  That's not addressed in this PR.

- Using an alternative physical layer for the same transport/control channel (i.e. Bluetooth PAN).  I didn't mention that in this PR but it could be included if there is interest.

louaybassbouss:
I will also add discovery and handshake to @mfoltzgoogle's interpretation of "guest more".
> For this PR, I interpreted "guest mode" to mean: "How to make this transport work for two endpoints that are connected to two distinct LANs and not directly routable."
As I understood from the [Guest mode documentation for Google Cast](https://developers.google.com/cast/docs/guest_mode) that BLE/Ultrasound/PIN are used for Discovery/pairing but I am not sure about transport. [WiFi-Direct](https://www.wi-fi.org/discover-wi-fi/wi-fi-direct) is one candidate that can be used for both "Discovery" and "Transport" in Guest mode.

mfoltzgoogle:
Discovery, pairing and transport across networks is bigger in scope than what I intend to address here.  My understanding of the webscreens charter [1] is that a complete guest mode stack is out of scope, so an amendment would be required to move forward with that in this repo.
 
[1] https://webscreens.github.io/cg-charter/

anssiko:
@mfoltzgoogle is correct. Quoting the [Scope of Work](https://webscreens.github.io/cg-charter/#scope-of-work):

>The following tasks are in scope for the work of the community group.
>
>1. The specification of a set of network protocols that allows one user agent to use the Presentation API to discover and present Web content on another user agent, when both user agents are connected to the same Local Area Network (LAN).

I suggest proposals that are out of scope for the current charter but that still have merit are tagged as "v2" issues.

louaybassbouss:
agree with @anssiko let's label this issue with v2 and move forward with the v1 features

mfoltzgoogle:
Marked the issue as "v2" and closing this pull request.  It should still be available in the archive when we revisit guest mode.

Issue #69, opened 2017-10-10
[Meta] Performance data for discovery protocols


mfoltzgoogle:
This meta-issue tracks the work to gather performance data on proposed discovery protocols.

Even though both SSDP and mDNS have been implemented in Chrome (and Mozilla?), for a variety of reasons, it can be tricky to use data from the field to compare performance of the discovery protocols themselves.  Therefore I propose running some experiments to gather accurate data.

- [x] Design experimental framework: what devices, what network, what software and what measurements.
- mDNS
  - [x] Issue #37: [mDNS] Get reliability data
  - [ ] Issue #38: [mDNS] Get network and power efficiency data
- SSDP
  - [x] Issue #68: [SSDP] Get reliability data
  - [ ] Issue #26: [SSDP] Get network and power efficiency data
- [ ] Collect data on latency of discovery (when devices enter/leave the LAN)

Separately from experiments, there may be some data we can share from Chrome internal metrics related to reliability.  I will look into it and report back.

louaybassbouss:
Maybe we can use these libraries to do some experiments:

-  [mdns](https://www.npmjs.com/package/mdns): A Node.js module the can be used to advertise and discover services. I think it uses Bonjour.
- [peer-ssdp](https://www.npmjs.com/package/peer-ssdp): A Node.js module provided by Fraunhofer FOKUS which implements the SSDP protocol (without UPnP).

mfoltzgoogle:
Based on TPAC discussion I'll take an action to share data on the success of SSDP and mDNS "in the field" and in particular how often one succeeds when the other fails.   If we break this down by platform, we can guess how often it is a device specific issue versus a networking issue.

mfoltzgoogle:
At the Berlin F2F I shared reliability data for SSDP and mDNS, and based on that we resolved to make mDNS mandatory for controllers and receivers, and make SSDP a non-mandatory alternative.

The issues to get networking and power benchmarking remain open, so leaving this meta-issue open.   It's possible additional data from benchmarking may change the tradeoffs here.

mfoltzgoogle:
We've already committed to mDNS in the current spec.

PR #71, opened 2017-10-16
Benchmark design for discovery protocols.

mfoltzgoogle:


mfoltzgoogle:
This is to collect data related to Issue #69 (Performance data for discovery protocols).

benchmarks/discovery.md
```diff
@@ -0,0 +1,148 @@
...
+household where controller devices, receiver devices, and routers are spread
+among different rooms.
+
+**NOTE:** If there is interest in evaluating these protocols in particularly
+adverse WiFi conditions, we can investigate using a routing proxy to simulate
+additional congestion and packet loss, or simply move devices further away.
+
+## SSDP specific
+
+* Interval between `M-SEARCH` requests.
```
louaybassbouss: Maybe we need also to consider the SSDP header `MX` which contains the maximum time in seconds (must be between 1 and 5 seconds) to delay `M-SEARCH` response. Also the `CACHE-CONTROL` header could be relevant.

benchmarks/discovery.md
```diff
@@ -0,0 +1,148 @@
...
+
+For numerical values, data will be collected for a given experimental condition
+N = 30 times and the 50%, 75%, and 95% values computed, as well as time series.
+
+One experimental condition (the control) will have no senders or responders to
+collect baseline data.
+
+# Notes
+
+Power consumption measurement only makes sense as a relative measurement, and
```
louaybassbouss: We can track battery level, number of discovery requests, number of found devices for both protocols mDNS and SSDP with the same number of senders and responders in the network. The experiment starts with fully charged smartphone and ends until the battery level reaches a pre-defined level e.g. 50%.

benchmarks/discovery.md
```diff
@@ -0,0 +1,148 @@
...
+
+For mDNS, we will use the [Avahi](https://www.avahi.org/) open source mDNS
+software that is available as a Debian package.
+
+For SSDP, we will either:
+
+1. Fork and customize
+[libupnp](http://pupnp.sourceforge.net/) for the [Open Screen SSDP discovery
+protocol](../ssdp.md).
+2. Write our own SSDP implementation from scratch, possibly based on the
```
louaybassbouss: Feel free to use this simple Node.js implementation https://github.com/fraunhoferfokus/peer-ssdp/blob/master/lib/peer-ssdp.js in the Benchmark responder test app which works also on Raspberry Pi. It implements only the SSDP protocol and not the whole UPnP Stack. For Android we can use this https://github.com/fraunhoferfokus/cordova-plugin-hbbtv/tree/master/src/android/ssdp class and use it in the Benchmark sender app.

louaybassbouss:
Thx @mfoltzgoogle I added some comments inline.

benchmarks/discovery.md
mfoltzgoogle: Done, added the Node.js as an alternative implementation for the SSDP stack here.  I also added a link to the Android plugin to the section at the end on mobile device benchmarking.

benchmarks/discovery.md
mfoltzgoogle: Good point - `MX` will definitely affect latency, added it as a variable.

`CACHE-CONTROL` does also impact the latency of discovery and the quantity of network traffic over a long time frame, since the value is suggested to be greater than 1800 seconds.  It could be interesting to vary it while making long term measurements over the course of hours (after discovery hits a steady state).  Added a note about steady state behavior as future work to the end.

benchmarks/discovery.md
mfoltzgoogle: That would be one approach, although I worry that the amount of battery consumed is so small relative to the other power sinks on a mobile device that it will get lost in the noise.  There are a lot of mobile-specific implementation techniques that can affect power like coalescing background tasks and/or radio wakeups for passive discovery mechanisms like these.

Added a note to look at battery levels as a first approximation; but this really requires some more research and I will file a task to follow up.

mfoltzgoogle:
Thanks @louaybassbouss.  I've updated the PR in response to your feedback.  Feel free to look again.

louaybassbouss:
LGTM

PR #76, opened 2017-10-22
Add Chrome OS to platform support

beaufortfrancois:
This CL simply adds Chrome OS to platform support.

R: @mfoltzgoogle


louaybassbouss:
thx @beaufortfrancois LGTM

PR #78, opened 2017-11-07
Fix links to issues

anssiko:
Short links do not seem to work in .md files, need to use absolute URLs instead.


anssiko:
Actually, relative URLs do work. Updated the PR accordingly.

mfoltzgoogle:
Thanks!

Issue #13, opened 2016-11-22
[Meta] Propose security mechanisms for Open Screen protocol


mfoltzgoogle:
This issue tracks proposed security mechanisms to authenticate the parties communicating via the open screen protocol. 

Before proceeding, I would like to solicit consensus about what is in scope and out of scope regarding authentication.  In particular, I proposed that the initial focus is on UA-to-UA authentication, versus authenticating users, applications, or the particular software running either UA.

These additional layers of authentication could potentially be written as protocol extensions to the base transport security, but would not be the initial focus of the proposals.

mwatson2:
I think the initial focus, as mentioned in the Charter, should be on ensuring that the _target device_ is the one selected by the user (i.e that there is no man-in-the-middle). Is this what you mean as UA-to-UA ?

mfoltzgoogle:
Yes.  What's out of scope for this initial RFP is securing the transport between the controlling and receiving UA, from either passive eavesdropping or MITM; this is explicitly called out in the charter.

Once that layer is specified, additional security layers can be built on top.  However, that work is not explicitly listed in the charter, and would likely be considered out of scope without a charter amendment.

tomoyukilabs:
Is it correct that [device paring in Mozilla proposal](https://wiki.mozilla.org/WebAPI/PresentationAPI:Protocol_Draft#Device_Pairing) can be one of the candidate mechanisms, for example?

chrisn:
I'm not sure I follow where secure transport is explicitly out of scope (I'm reading item 2 under Scope of Work and item 1 under Deliverables in the charter).

mfoltzgoogle:
@tomoyukilabs Yes the Mozilla proposal is certainly a candidate and they are invited to write it up and submit it here.

@chrisn That was a mistake, thank you for catching it :) I corrected my earlier comment.

anssiko:
Mozilla's [Device Pairing](https://wiki.mozilla.org/WebAPI/PresentationAPI:Protocol_Draft#Device_Pairing) proposal by @schien is being discussed at F2F, minutes:

https://www.w3.org/2017/11/07-webscreens-minutes.html#x14

mfoltzgoogle:
From https://www.w3.org/2017/11/07-webscreens-minutes.html#x16:
- ACTION: @schien to contribute J-PAKE proposal to repo
- @mfoltzgoogle ACTIONs:
  - review Internet-Draft for ecjpake
[...]

PR #85, opened 2018-02-26
Add a document about details on J-PAKE

tomoyukilabs:
This PR adds a document that summarizes the J-PAKE protocol described in [RFC 8236](https://datatracker.ietf.org/doc/rfc8236/), including:

- what types of messages must be exchanged between two parties
- two possible architectures: J-PAKE over a dedicated channel, and TLS with ECJ-PAKE
- open source implementations

I'd be happy if this would be useful for our discussion.

### DIFF
--- a/j-pake.md
+++ b/j-pake.md
@@ -1,0 +1,246 @@
+# J-PAKE: Password-Authenticated Key Exchange by Juggling
+
+Password-Authenticated Key Exchange by Juggling (J-PAKE) is a protocol that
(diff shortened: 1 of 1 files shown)


anssiko:
Thanks @tomoyukilabs, LGTM! To be discussed in the context of https://github.com/webscreens/openscreenprotocol/issues/13 at https://www.w3.org/wiki/Second_Screen/Meetings/May_2018_F2F

PR #90, opened 2018-05-19
add schemes.md

louaybassbouss:
Add explanation for usage of schemes in OSP (and Presentation API) to support multiple receiver UAs from different vendors

PR #95, opened 2018-06-12
Update schemes.md

mfoltzgoogle:
This addresses comments from PR #90.

Specifically, it makes some copy editing changes to rewrap the text, use 'URL' instead of 'url', and 'presentation screen' instead of 'receiving UA'.  

It also adds some new content to explain how presentation screens that support multiple schemes may be handled.

Finally, it adds links to open issues to comment on interop, define hbbtv: parameters more concretely, and links this document from the control protocol document where URLs are involved.

### DIFF
--- a/control_protocol.md
+++ b/control_protocol.md
@@ -313,6 +313,10 @@ receiver.
 
 **TODO**: Add messages for this based on conclusion to [Issue #21](issues/21).
 
+Note that the URLs sent by the Presentation Availability Request may contain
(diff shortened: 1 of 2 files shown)


mfoltzgoogle:
@louaybassbouss Would you like to review this? If I don't hear back soon I will merge, but feel free to send feedback afterwards if you want.

louaybassbouss:
Thanks a lot @mfoltzgoogle for your update LGTM I will merge this PR

Issue #57, opened 2017-10-05
[SSDP] Update proposed use of SSDP to specifically prevent SSDP amplification attacks


mfoltzgoogle:
Background: https://blog.cloudflare.com/ssdp-100gbps/

mfoltzgoogle:
Closed by PR #96.

PR #96, opened 2018-06-12
Prevent SSDP amplification attacks

mfoltzgoogle:
Addresses Issue #57: [SSDP] Update proposed use of SSDP to specifically prevent SSDP amplification attacks.

This requires mitigations which will prevent an attacker from using an Open Screen Protocol implementation to target an external IP address with SSDP responses.


mfoltzgoogle:
Going ahead and merging, as this is relatively straightforward. Feel free to add comments afterwards.

PR #97, opened 2018-08-24
Outline for a combined specification document.

mfoltzgoogle:
Hello,

To incorporate existing material and add new material from our F2F meeting in May, I'd like to start a combined document that I think will be easier to iterate on than the scattered set of existing documents.

Most of the material will be based on those documents or meeting minutes; however, I'd also like to make the technical descriptions more "spec like" using specific procedures and steps, and conformance terms.

Ideally, this will be substantially complete (as far as incorporating existing material) in advance of TPAC.

This PR adds three files:
- Spec outline using [Bikeshed](https://tabatkins.github.io/bikeshed/)
- Contribution mechanics (since we don't have one for SSCG)
- Simple Makefile to regenerate spec.

I will add a meta-issue to fill in sections after this is complete, and issues for any sections that don't already have them.

Regarding Bikeshed:

This is now in use across more groups, and I feel will be easier to use, as all of our existing material is already in Markdown.

Since our group doesn't have it's own boilerplate in the Bikeshed repo, I'm making do with the default boilerplate and added the W3C copyright.  I can followup with Tab Atkins if it would be simpler to do all of our custom boilerplate in-document, or go through the process of setting up group boilerplate.

PTAL @louaybassbouss @chrisn @tidoust @pthatcherg @anssiko

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3,7 +3,7 @@ Title: Open Screen Protocol
 Shortname: openscreenprotocol
 Level: 1
 Status: w3c/ED
-URL: https://webscreens.github.io/openscreenprotocol/
+ED: https://webscreens.github.io/openscreenprotocol/
 Canonical URL: ED
 Editor: Mark Foltz, Google, https://github.com/mfoltzgoogle, w3cid 68454
 Repository: webscreens/openscreenprotocol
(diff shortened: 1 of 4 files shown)


mfoltzgoogle:
Thanks @louaybassbouss and @chrisn.

Before landing, I can ensure each TODO has a corresponding issue and assign them through GitHub.  I filed a meta-issue #98 to cover all of the work remaining to finish a first draft of this document.

@tidoust or @anssiko, can you to take a look to make sure the initial boilerplate looks ok?  If I don't hear in another day or so, I'll land and we can address in a followup PR.

mfoltzgoogle:
Going ahead and landing to keep things moving along.  Feel free add comments to this thread when you have a chance.

tidoust:
See PR #104 for suggested adjustments to the boilerplate text. Copyright for CG documents is not the same as the usual W3C copyright in particular, because CG specs are published under the [CLA](https://www.w3.org/community/about/agreements/cla/).

index.bs
```diff
@@ -0,0 +1,99 @@
...
+
+Receiver Discovery {#discovery}
+===============================
+
+Issue(100): Incorporate material from the [mDNS](mdns.md) document.
+
+Transport Establishment {#transport}
+====================================
+
+Issue(101): Incorporate material from the \[QUIC](quic.md) document.
```
pthatcherg: What is the difference between [foo] and \[foo] ?  mDNS above has no "\" while QUIC and J-PAKE here do.

mfoltzgoogle: I think you're asking about the backslash.  This is escaping to prevent bikeshed from treating these as text macros.  I probably added them when the bikeshed compiler complained.

https://tabatkins.github.io/bikeshed/#text-macros

PR #119, opened 2018-11-09
Specify use of mDNS, QUIC, and CBOR

pthatcherg:


index.bs
```diff
@@ -232,27 +232,117 @@ Non-Functional Requirements {#requirements-non-functional}
...
+  [RFC 8122 section 5](https://tools.ietf.org/html/rfc8122#section-5),
+  excluding the "fingerprint:" prefix and including the hash function, space, and
+  hex-encoded fingerprint.  The fingerprint value also functions as an ID for the
+  endpoint. All endpoints must support the following hash functions: "sha-256",
+  "sha-512".  Endpoints must not support the following hash functions: "md2",
+  "md5". 
+- key 'time" with value of the last time the endpoint updated its
+  metadata.  The time value is respesented as zero or a positive integer.  The
+  clock and timescale are chosen by the advertising endpoint, but are not
+  important, except that the clock must be montonic.  A larger value is an
```
tidoust:
```suggestion
  important, except that the clock must be monotonic.  A larger value is an
```

index.bs
```diff
@@ -232,27 +232,117 @@ Non-Functional Requirements {#requirements-non-functional}
...
+  indication to discovering clients that they should reconnect to the adverisiting
```
tidoust:
```suggestion
  indication to discovering clients that they should reconnect to the advertising
```

index.bs
```diff
@@ -232,27 +232,117 @@ Non-Functional Requirements {#requirements-non-functional}
...
+Transport and metadata discovery with QUIC {#transport}
 ====================================
 
-Issue(101): Incorporate material from the \[QUIC](quic.md) document.
+If a discovering endpoint wants to connect to an advertising endpoint, or to
+learn further metdata about it, it initiates a
+[QUIC](https://tools.ietf.org/html/draft-ietf-quic-transport-16) connection to
+the IP and port of the mDNS entry.  Prior to authentication, some message may be
+exchanged (such as further metadata), but such info should be treated as
+unathenticated (such as indicating to a user that a display name of an
```
tidoust:
```suggestion
unauthenticated (such as indicating to a user that a display name of an
```

index.bs
```diff
@@ -232,27 +232,117 @@ Non-Functional Requirements {#requirements-non-functional}
...
+unathenticated endpoint is unathenticated).
```
tidoust:
```suggestion
unauthenticated endpoint is unauthenticated).  
```

index.bs
```diff
@@ -232,27 +232,117 @@ Non-Functional Requirements {#requirements-non-functional}
...
-Issue(101): Incorporate material from the \[QUIC](quic.md) document.
+If a discovering endpoint wants to connect to an advertising endpoint, or to
+learn further metdata about it, it initiates a
+[QUIC](https://tools.ietf.org/html/draft-ietf-quic-transport-16) connection to
+the IP and port of the mDNS entry.  Prior to authentication, some message may be
+exchanged (such as further metadata), but such info should be treated as
+unathenticated (such as indicating to a user that a display name of an
+unathenticated endpoint is unathenticated).  
+
+To learn further metadata, the an endpoint may send a endpoint-info-request
```
tidoust:
```suggestion
To learn further metadata, an endpoint may send a endpoint-info-request
```

index.bs
```diff
@@ -232,27 +232,117 @@ Non-Functional Requirements {#requirements-non-functional}
...
+message (see Appendix A) and receive back a endpoint-info-response message.  The
+messages may contain the following information with the following meaning.
```
tidoust:
```suggestion
messages may contain the following information with the following meaning:
```

index.bs
```diff
@@ -232,27 +232,117 @@ Non-Functional Requirements {#requirements-non-functional}
...
+unathenticated endpoint is unathenticated).  
+
+To learn further metadata, the an endpoint may send a endpoint-info-request
+message (see Appendix A) and receive back a endpoint-info-response message.  The
+messages may contain the following information with the following meaning. 
+- display-name: The display name of the responding endpoint intended to be
+  displayed to a user by the requesting endpoint.  If the responding endpoint is
+  not yet authenticated, the requesting endpoint should make UI affordance for
+  indicating to the user that the display name is not yet authenticated.  If the
+  responding endpoint changes it display name, the requesting endpoing should make
```
tidoust:
```suggestion
  responding endpoint changes its display name, the requesting endpoint should make
```

index.bs
```diff
@@ -232,27 +232,117 @@ Non-Functional Requirements {#requirements-non-functional}
...
+  UI affordance for indicating to the user that the display name has changed. 
+- model-name: The model name of the device (if it is a device), used mainly for
+  debugging purpuses, but may be displayed to the user of the requesting endpoint.
```
tidoust:
```suggestion
  debugging purposes, but may be displayed to the user of the requesting endpoint.
```

index.bs
```diff
@@ -232,27 +232,117 @@ Non-Functional Requirements {#requirements-non-functional}
...
+- key "fp" with value of the certificate fingerpint of the endpoint.  The format
+  of the fingerprint is defined by 
+  [RFC 8122 section 5](https://tools.ietf.org/html/rfc8122#section-5),
+  excluding the "fingerprint:" prefix and including the hash function, space, and
+  hex-encoded fingerprint.  The fingerprint value also functions as an ID for the
+  endpoint. All endpoints must support the following hash functions: "sha-256",
+  "sha-512".  Endpoints must not support the following hash functions: "md2",
+  "md5". 
+- key 'time" with value of the last time the endpoint updated its
+  metadata.  The time value is respesented as zero or a positive integer.  The
```
tidoust:
```suggestion
  metadata.  The time value is represented as zero or a positive integer.  The
```

index.bs
```diff
@@ -232,27 +232,130 @@ Non-Functional Requirements {#requirements-non-functional}
...
+
+- key "fp" with value of the certificate fingerpint of the endpoint.  The format
+  of the fingerprint is defined by [RFC 8122 section
+  5](https://tools.ietf.org/html/rfc8122#section-5), excluding the
+  "fingerprint:" prefix and including the hash function, space, and hex-encoded
+  fingerprint.  The fingerprint value also functions as an ID for the endpoint.
+  All endpoints must support the following hash functions: "sha-256", "sha-512".
+  Endpoints must not support the following hash functions: "md2", "md5". 
+
+- key 'time" with value of the last time the endpoint updated its
```
```suggestion
- key "time" with value of the last time the endpoint updated its
```

index.bs
```diff
@@ -232,27 +232,130 @@ Non-Functional Requirements {#requirements-non-functional}
...
+  metadata.  The time value is represented as zero or a positive integer.  The
+  clock and timescale are chosen by the advertising endpoint, but are not
```
btolsch: nit: Would it make sense for this to just be specified as an arbitrary value that is _at least_ incremented whenever the metadata is updated, since it can't really be used as a time?

index.bs
```diff
@@ -232,27 +232,130 @@ Non-Functional Requirements {#requirements-non-functional}
...
 
-Transport Establishment {#transport}
+Transport and metadata discovery with QUIC {#transport}
 ====================================
 
-Issue(101): Incorporate material from the \[QUIC](quic.md) document.
+If a discovering endpoint wants to connect to an advertising endpoint, or to
+learn further metdata about it, it initiates a
+[QUIC](https://tools.ietf.org/html/draft-ietf-quic-transport-16) connection to
+the IP and port of the mDNS entry.  Prior to authentication, some message may be
```
```suggestion
the IP and port from the SRV record.  Prior to authentication, some message may be
```

index.bs
```diff
@@ -232,27 +232,130 @@ Non-Functional Requirements {#requirements-non-functional}
...
+exchanged (such as further metadata), but such info should be treated as
+unauthenticated (such as indicating to a user that a display name of an
+unauthenticated endpoint is unauthenticated).  
+
+To learn further metadata, an endpoint may send a endpoint-info-request
```
```suggestion
To learn further metadata, an endpoint may send an endpoint-info-request
```

index.bs
```diff
@@ -232,27 +232,130 @@ Non-Functional Requirements {#requirements-non-functional}
...
+message (see Appendix A) and receive back a endpoint-info-response message.  The
```
```suggestion
message (see Appendix A) and receive back an endpoint-info-response message.  The
```

index.bs
```diff
@@ -279,5 +382,56 @@ Threat model {#security-threat}
...
+<pre>
+; type key 10
+endpoint-info-request = {
+  request
+}
+
+; type key 11
+endpoint-info-response = {
+  response
+  1: endpoint-info;
```
```suggestion
  1: endpoint-info ; endpoint-info
```

index.bs
```diff
@@ -279,5 +382,56 @@ Threat model {#security-threat}
...
+}
+
+endpoint-info = {
+  0: text ; friendly-name
+  1: text ; model-name
+  ; ...
+}
+
+; type key 12
+status-request = {
```
```suggestion
endpoint-status-request = {
```

index.bs
```diff
@@ -279,5 +382,56 @@ Threat model {#security-threat}
...
+  request
+  ? 1: status ; status
+}
+
+; type key 13
+status-response = {
```
```suggestion
endpoint-status-response = {
```

index.bs
```diff
@@ -279,5 +382,56 @@ Threat model {#security-threat}
 Mitigations and defenses {#security-defenses}
 ---------------------------------------------
 
+Appending A: Messages
+=====================
+The following messages are defined with [CDDL](https://datatracker.ietf.org/doc/draft-ietf-cbor-cddl/). When integer keys are used, a comment is appended to the line to indicate the name of the field. Each root message (one that can be put into a QUIC stream without being inclosed by another message) has a comment indicating the message type key.
```
btolsch: Suggest line breaks here and below (I would use GitHub's suggest here as well but there's no column count in the comment editor).

index.bs
```diff
@@ -232,27 +232,130 @@ Non-Functional Requirements {#requirements-non-functional}
...
+   integer (see Appendix A for type keys) 
+
+2. The message length encoded as a variable-length integer 
+
+3. The messge encoded as CBOR (whose length must match the prefixed length)
+
+Future message type keys may indicate the use of custom message formats that are
+not CBOR, or use of CBOR without the length prefix, so receiving endpoints must
+disregard messages with type keys that it does not recognize and must not assume
+that they are serialized using CBOR nor that they have a length prefix.
```
btolsch: Does this mean then that the endpoint must close the QUIC stream (possibly RST_STREAM + application error)?  If it can't assume there's a length prefix (which makes sense, given we may omit it for some types), it can't skip that message and then continue reading.

If this isn't desirable, one possibility would be to send [length, message] as a CBOR array (and just [message] when there's no length prefix.  In that case, the CBOR parser would still be able to step over it.

index.bs
```diff
@@ -232,27 +232,130 @@ Non-Functional Requirements {#requirements-non-functional}
...
+extended to include additional information not defined in this spec.  If done
+adhoc by applications and not in future specs, keys should be chosen to avoid
+collision, such as by choosing large integers or long strings.  
+
+Messages delivery using CBOR and QUIC streams {#control}
+========================================================
+
+Messages are serialized using [CBOR](https://tools.ietf.org/html/rfc7049).  To
+send a group of messages in order, that group of messages must be sent in one
+QUIC stream.  Different groups of messages which are ordered within the group
```
btolsch: This sentence didn't really make sense to me, possibly because of the double meaning of "group".  Can you give an example or clarify further?

btolsch:
Wow, so apparently making everything a "single comment" had disastrous consequences.  I'm new to reviewing on GitHub, so sorry for the spam!

index.bs
```diff
@@ -232,27 +232,130 @@ Non-Functional Requirements {#requirements-non-functional}
...
+If a discovering endpoint (acting as a QUIC client) wishes to receive messages
+from an advertising endpoint (acting as a QUIC server) or an advertising
+endpoint (acting as a QUIC server) wishes to send message to a discovering
+endpoint (acting as a client endpoint), it may wish to keep the QUIC connection
+alive.  Once neither side needs to keep the connection alive for the purposes of
+sending or receiving messages, the connection should be closed.  In order to
+keep a QUIC connection alive, an endpoint may send an endpoint-status-request
+message, and any endpoint that receives an endpoint-status-request message
+should send an endpoint-status-response message.  Such messages should be sent
+every 25 seconds.  If, however, other mechanisms or traffic (such as other
```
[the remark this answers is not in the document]
btolsch: I just saw that under the `PING` frame description, the answer is basically "yes, but..." It sounds like `idle_timeout` would only be useful if it would lower our 25 second timer.

index.bs
pthatcherg: Good idea.  I updated it thus.

index.bs
pthatcherg: I updated it to make it more clear.

index.bs
pthatcherg: Ah, good point.  I was mainly thinking of this for messages where there would be only one message in a stream, such as audio or video frames.  But even then, you have a good point about whether the receiving endpoint should reset the stream or not.  

Perhaps I should change this idea and instead require apps to check for capabilities of receiving new message types before sending them.  And we should always tear down the entire connection if an unknown type key is received.

pthatcherg: OK, I did that: close the connection with a specific error code.

index.bs
pthatcherg: Done

index.bs
```diff
@@ -232,27 +232,130 @@ Non-Functional Requirements {#requirements-non-functional}
...
+not be closed because of idle timeout unless no such mechanisms or traffic has
+been received from the remote side for 35 seconds (allowing for a keep-alive
+frequency of 25 seconds and message delivery delay of 10 seconds).
+
+If a client endpoint wishes to send messages to a server endpoint, the client
+endpoint can connect to the server endpoint "on demand": it does not need to
+keep the connection alive.
+
+The endpoint-info-response message and endpoint-status-response messages may be
+extended to include additional information not defined in this spec.  If done
```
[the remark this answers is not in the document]
pthatcherg: OK, I made it explicit.

index.bs
```diff
@@ -232,27 +232,130 @@ Non-Functional Requirements {#requirements-non-functional}
...
  
 
-Receiver Discovery {#discovery}
+Discovery with mDNS {#discovery}
 ===============================
 
-Issue(100): Incorporate material from the [mDNS](mdns.md) document.
+Endpoints may discover one another using
+[mDNS](https://tools.ietf.org/html/rfc6762) and
+[DNS-SD](https://tools.ietf.org/html/rfc6763).  To do so, endpoints must use the
```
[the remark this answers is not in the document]
pthatcherg: OK, I change it to "using DNS-SD over mDNS"

index.bs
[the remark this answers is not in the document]
pthatcherg: That's a good point about the idle_timeout transport parameter.  Perhaps we should just say "send more frequently than idle_timeout" and not specify 25s (since the QUIC spec recommends 15-30 seconds).

mfoltzgoogle:
@pthatcherg Would it be possible to include generated HTML alongside this pull request?  It helps determine what the end result of formatting looks like.  You would need to [install Bikeshed](https://tabatkins.github.io/bikeshed/#installing) then run the Makefile in the repo.

Not a big deal as this can be done later as well.

index.bs
```diff
@@ -232,27 +232,137 @@ Non-Functional Requirements {#requirements-non-functional}
...
+transport parameter (see section 18 of
+[QUIC](https://tools.ietf.org/html/draft-ietf-quic-transport-16)) and QUIC PING
+frames should not be used.  An idle_timeout transport parameter of 25 seconds is
+recommended.  If other mechanisms or traffic (such as other requests and
+responses) are serving the purpose of keeping the connection alive,
+endpoint-status-requests messages should be delayed or not used at all.
+
+If a client endpoint wishes to send messages to a server endpoint, the client
+endpoint can connect to the server endpoint "on demand"; it does not need to
+keep the connection alive.
```
mfoltzgoogle: Ah, so the keepalive is always initiated from the QUIC server?

index.bs
```diff
@@ -232,27 +232,137 @@ Non-Functional Requirements {#requirements-non-functional}
...
+
+The endpoint-info-response message and endpoint-status-response
+messages may be extended to include additional information not defined
+in this spec.  If done ad hoc by applications and not in future specs,
+keys should be chosen to avoid collision, such as by choosing large
+integers or long strings.  Endpoints must ignore keys in the
+endpoint-info-message that it does not understand to allow endpoints
+to easily extend this message.
```
mfoltzgoogle: We discussed at TPAC defining tag/key ranges for extensions, and possibly registering some of them with IANA.

I will take a TODO to file a GitHub issue to spin off that discussion and we can incorporate the results here.

index.bs
```diff
@@ -232,27 +232,137 @@ Non-Functional Requirements {#requirements-non-functional}
...
+
+Messages delivery using CBOR and QUIC streams {#control}
+========================================================
+
+Messages are serialized using [CBOR](https://tools.ietf.org/html/rfc7049).  To
+send a group of messages in order, that group of messages must be sent in one
+QUIC stream.  Different groups of messages (with no ordering dependency
```
mfoltzgoogle:
```suggestion
QUIC stream.  Independent groups of messages (with no ordering dependency
```

index.bs
```diff
@@ -232,27 +232,137 @@ Non-Functional Requirements {#requirements-non-functional}
...
+across groups) should be sent in different QUIC streams.  In order to put
+multiple CBOR-serialiezd messages into the the same QUIC stream, the following
```
mfoltzgoogle:
```suggestion
multiple CBOR-serialized messages into the the same QUIC stream, the following
```

index.bs
```diff
@@ -232,27 +232,137 @@ Non-Functional Requirements {#requirements-non-functional}
...
+is used.
+
+For each message, the sender must write to the QUIC stream the following: 
+1. A type key representing the type of the message, encoded as a variable-length
+   integer (see Appendix A for type keys) 
+
+2. The message length encoded as a variable-length integer 
+
+3. The messge encoded as CBOR (whose length must match the prefixed length)
```
mfoltzgoogle:
```suggestion
3. The messge encoded as CBOR (whose length must match the value in step 2)
```

index.bs
```diff
@@ -232,27 +232,137 @@ Non-Functional Requirements {#requirements-non-functional}
...
+
+If an endpoint receives a message for which it does not recognize a
+type key, it must close the QUIC connection with an application error
+code of 2 and should include the unknown type key in the reason phrase
+(see [QUIC](https://tools.ietf.org/html/draft-ietf-quic-transport-16)
+sections 19.4).
+
+Variable-length integers are encoded in the same format as defined by [QUIC
```
mfoltzgoogle: Instead of a CBOR varint?  Since the implementation has to parse CBOR anyway, it might be easier to use CBOR here too.

index.bs
```diff
@@ -232,27 +232,137 @@ Non-Functional Requirements {#requirements-non-functional}
...
+
+Advertising endpoints must include DNS TXT records with the following
+keys and values:
+
+- key "fp" with value of the certificate fingerpint of the advertising endpoint.
+  The format of the fingerprint is defined by [RFC 8122 section
+  5](https://tools.ietf.org/html/rfc8122#section-5), excluding the
+  "fingerprint:" prefix and including the hash function, space, and hex-encoded
+  fingerprint.  The fingerprint value also functions as an ID for the endpoint.
+  All endpoints must support the following hash functions: "sha-256", "sha-512".
```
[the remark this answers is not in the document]
pthatcherg: Added the TODO

index.bs
```diff
@@ -232,27 +232,137 @@ Non-Functional Requirements {#requirements-non-functional}
...
-Issue(100): Incorporate material from the [mDNS](mdns.md) document.
-
-Transport Establishment {#transport}
-====================================
-
-Issue(101): Incorporate material from the \[QUIC](quic.md) document.
+Endpoints may discover one another using
+[DNS-SD](https://tools.ietf.org/html/rfc6763) over
+[mDNS](https://tools.ietf.org/html/rfc6762).  To do so, endpoints must
+use the service name "_openscreen._udp.local".
```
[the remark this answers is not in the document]
pthatcherg: Can't an endpoint choose any port?  What restrictions would we impose?

index.bs
```diff
@@ -232,27 +232,137 @@ Non-Functional Requirements {#requirements-non-functional}
...
+  fingerprint.  The fingerprint value also functions as an ID for the endpoint.
+  All endpoints must support the following hash functions: "sha-256", "sha-512".
+  Endpoints must not support the following hash functions: "md2", "md5". 
+
+- key "time" with arbitrary unsigned integer value.  To indicate that
+  metadata has changed, the advertising endpoint must update it to a
+  greater value.  This signals to the discovering endpoint that it
+  should connect to the advertising endpoint to discover updated
+  metadata.
+
```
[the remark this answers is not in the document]
pthatcherg: I added a TODO

index.bs
```diff
@@ -232,27 +232,137 @@ Non-Functional Requirements {#requirements-non-functional}
...
+- key "fp" with value of the certificate fingerpint of the advertising endpoint.
+  The format of the fingerprint is defined by [RFC 8122 section
+  5](https://tools.ietf.org/html/rfc8122#section-5), excluding the
+  "fingerprint:" prefix and including the hash function, space, and hex-encoded
+  fingerprint.  The fingerprint value also functions as an ID for the endpoint.
+  All endpoints must support the following hash functions: "sha-256", "sha-512".
+  Endpoints must not support the following hash functions: "md2", "md5". 
+
+- key "time" with arbitrary unsigned integer value.  To indicate that
```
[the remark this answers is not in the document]
pthatcherg: It's not a timestamp.  It's more like a metadata version that's incremented.  We could go with "mv" for "metadata version" or something like that.

index.bs
pthatcherg: I think the CBOR varints are too big.  I went through many different designs and options for this, and this was the best mix of small and easy to implement.  If you read the spec I linked to, you'll see it's very easy to implement.  And if you work through the CBOR varints, I think you'll find that it will use up an extra byte per int (2 bytes per message) most of the time.

index.bs
```diff
@@ -232,27 +232,137 @@ Non-Functional Requirements {#requirements-non-functional}
...
+  metadata.
+ 
+An endpoint that advertises this service name must implement the
+transport and metadata discovery mechanism with QUIC described below.
+If a future version of this protocol uses mDNS but breaks
+compatibility with that mechanism, it should change the service name
+to a new value indicating a new mechanism for metadata discovery
+mechanism.  If future versions of this protocol or extensions to this
+protocol are compatible with the metadata discovery mechanism, they
+can use it to indicate support for future versions and extensions.
```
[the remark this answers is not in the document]
pthatcherg: No, what I meant was if we have new things added (like streaming), as long as we continue to use QUIC and CBOR to learn that one does the new thing, one does not need to change the service name.  But if one change away from QUIC or CBOR, one would need to change the service name because QUIC and CBOR are the things used to learn about new things.

It's a bit tricky to explain.  Ideas for how to improve are welcome.

index.bs
```diff
@@ -232,27 +232,137 @@ Non-Functional Requirements {#requirements-non-functional}
...
+Discovery with mDNS {#discovery}
 ===============================
 
-Issue(100): Incorporate material from the [mDNS](mdns.md) document.
-
-Transport Establishment {#transport}
-====================================
-
-Issue(101): Incorporate material from the \[QUIC](quic.md) document.
+Endpoints may discover one another using
```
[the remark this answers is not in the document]
pthatcherg: I just mean an implementation of the protocol that wants to talk to other implementations of the protocol.  I've seen specs use words like "endpoint", "agent", and "implementation".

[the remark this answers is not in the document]
pthatcherg: When there's a particular role (advertiser, discoverer, client, server, controller, receiver, etc), I use that.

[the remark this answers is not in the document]
pthatcherg: OK, I defined "agent" and went with that.  It's closer to "user agent" and is what the ICE spec uses.

index.bs
```diff
@@ -232,27 +232,137 @@ Non-Functional Requirements {#requirements-non-functional}
...
+to a new value indicating a new mechanism for metadata discovery
+mechanism.  If future versions of this protocol or extensions to this
+protocol are compatible with the metadata discovery mechanism, they
+can use it to indicate support for future versions and extensions.
+
+
+Transport and metadata discovery with QUIC {#transport}
+=======================================================
+
+If a discovering endpoint wants to connect to an advertising endpoint, or to
```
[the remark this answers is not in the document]
pthatcherg: Done

index.bs
```diff
@@ -232,27 +232,137 @@ Non-Functional Requirements {#requirements-non-functional}
...
+learn further metdata about it, it initiates a
+[QUIC](https://tools.ietf.org/html/draft-ietf-quic-transport-16) connection to
+the IP and port from the SRV record.  Prior to authentication, some message may be
+exchanged (such as further metadata), but such info should be treated as
+unauthenticated (such as indicating to a user that a display name of an
+unauthenticated endpoint is unauthenticated).
```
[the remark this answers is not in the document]
pthatcherg: Done

index.bs
```diff
@@ -232,27 +232,137 @@ Non-Functional Requirements {#requirements-non-functional}
...
+the IP and port from the SRV record.  Prior to authentication, some message may be
+exchanged (such as further metadata), but such info should be treated as
+unauthenticated (such as indicating to a user that a display name of an
+unauthenticated endpoint is unauthenticated).  
+
+To learn further metadata, an endpoint may send an endpoint-info-request
+message (see Appendix A) and receive back an endpoint-info-response message.  The
+messages may contain the following information with the following meaning:
+
+- display-name: The display name of the responding endpoint intended to be
```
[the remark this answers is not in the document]
pthatcherg: Appendix A has the CDDL.  I don't know what you mean by "legal syntax".

index.bs
```diff
@@ -232,27 +232,137 @@ Non-Functional Requirements {#requirements-non-functional}
...
+  displayed to a user by the requesting endpoint.  If the responding endpoint is
+  not yet authenticated, the requesting endpoint should make UI affordance for
+  indicating to the user that the display name is not yet authenticated.  If the
```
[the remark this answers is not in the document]
pthatcherg: Yes, done

index.bs
```diff
@@ -232,27 +232,137 @@ Non-Functional Requirements {#requirements-non-functional}
...
+  responding endpoint changes its display name, the requesting endpoint should
+  make UI affordance for indicating to the user that the display name has
+  changed.
```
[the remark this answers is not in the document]
pthatcherg: To change a display name, an advertising agent (I'll go with that term) would need to change the instance name, the TXT "mv" (I'll go with that term).  A listening agent would then request the new name to verify it.  But it could do that over the existing QUIC connection.  It wouldn't need to drop the existing QUIC connection.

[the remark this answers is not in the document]
pthatcherg: I suppose we could optimize this by just allowing the adveritising agent to push new messages (events) that such info had been updated without the listening agent requesting it.

index.bs
```diff
@@ -232,27 +232,137 @@ Non-Functional Requirements {#requirements-non-functional}
...
+
+- display-name: The display name of the responding endpoint intended to be
+  displayed to a user by the requesting endpoint.  If the responding endpoint is
+  not yet authenticated, the requesting endpoint should make UI affordance for
+  indicating to the user that the display name is not yet authenticated.  If the
+  responding endpoint changes its display name, the requesting endpoint should
+  make UI affordance for indicating to the user that the display name has
+  changed. 
+
+- model-name: The model name of the device (if it is a device), used mainly for
```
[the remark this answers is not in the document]
pthatcherg: I made it clear it's optional.

index.bs
```diff
@@ -232,27 +232,137 @@ Non-Functional Requirements {#requirements-non-functional}
...
+  debugging purposes, but may be displayed to the user of the requesting
+  endpoint.
+
+<!-- TODO: Add device type and/or capabilities -->
+
+If a discovering endpoint (acting as a QUIC client) wishes to receive messages
```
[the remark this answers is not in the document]
pthatcherg: OK, done.

index.bs
```diff
@@ -232,27 +232,137 @@ Non-Functional Requirements {#requirements-non-functional}
...
+from an advertising endpoint (acting as a QUIC server) or an advertising
+endpoint (acting as a QUIC server) wishes to send message to a discovering
+endpoint (acting as a client endpoint), it may wish to keep the QUIC connection
+alive.  Once neither side needs to keep the connection alive for the purposes of
+sending or receiving messages, the connection should be closed with an error
+code of 1.  In order to keep a QUIC connection alive, an endpoint may send an
+endpoint-status-request message, and any endpoint that receives an
+endpoint-status-request message should send an endpoint-status-response message.
```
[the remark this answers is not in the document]
pthatcherg: Yes, we could do that.  Or we could have an endpoint-status-event message that is like the response but without the request.  It's an "I'm still here; here's my info" message.  That would still allow piggybacking of info and have the same number of messages.

index.bs
pthatcherg: No, the client may send them to continue receiving message from the server.

index.bs
```diff
@@ -232,27 +232,137 @@ Non-Functional Requirements {#requirements-non-functional}
...
+Such messages should be sent more frequently than the QUIC idle_timeout
+transport parameter (see section 18 of
+[QUIC](https://tools.ietf.org/html/draft-ietf-quic-transport-16)) and QUIC PING
+frames should not be used.  An idle_timeout transport parameter of 25 seconds is
+recommended.  If other mechanisms or traffic (such as other requests and
```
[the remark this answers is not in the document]
pthatcherg: OK, I added the concept of a timer.  See what you think.

index.bs
```diff
@@ -232,27 +232,137 @@ Non-Functional Requirements {#requirements-non-functional}
...
+
+<!-- TODO: Add device type and/or capabilities -->
+
+If a discovering endpoint (acting as a QUIC client) wishes to receive messages
+from an advertising endpoint (acting as a QUIC server) or an advertising
+endpoint (acting as a QUIC server) wishes to send message to a discovering
+endpoint (acting as a client endpoint), it may wish to keep the QUIC connection
+alive.  Once neither side needs to keep the connection alive for the purposes of
+sending or receiving messages, the connection should be closed with an error
+code of 1.  In order to keep a QUIC connection alive, an endpoint may send an
```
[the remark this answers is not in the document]
pthatcherg: We could just follow the rules that QUIC connections already have around idle timeouts.  Then as long as the other side is sending us something (and with the current rules, we send something back), the QUIC connection will stay alive.  

Or we could make an explicit "please don't close for X seconds" in the message.

index.bs
```diff
@@ -232,27 +232,137 @@ Non-Functional Requirements {#requirements-non-functional}
...
+
+If a discovering endpoint wants to connect to an advertising endpoint, or to
+learn further metdata about it, it initiates a
+[QUIC](https://tools.ietf.org/html/draft-ietf-quic-transport-16) connection to
+the IP and port from the SRV record.  Prior to authentication, some message may be
+exchanged (such as further metadata), but such info should be treated as
+unauthenticated (such as indicating to a user that a display name of an
+unauthenticated endpoint is unauthenticated).  
+
+To learn further metadata, an endpoint may send an endpoint-info-request
```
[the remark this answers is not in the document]
pthatcherg: Francois volunteered to do that :).

pthatcherg:
I'll install bikeshed and run it tomorrow. 

In the meantime, there are a few things remaining for discussion but I think I've resolved most of them.

index.bs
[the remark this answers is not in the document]
mfoltzgoogle: Well ports below 1024 are traditionally reserved for system services/daemons.  And if we don't want a future service conflicting with a port number chosen by some implementation, we could attempt to reserve one through IANA:

https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml

Probably better to resolve in a GitHub issue as it's not obvious what port(s) to assign for now.

index.bs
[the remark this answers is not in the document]
mfoltzgoogle: My suggestion for the last two sentences:

> Future extensions to this QUIC-based protocol can use the same metadata discovery process to indicate support for those extensions, through a capabilities mechanism to be determined.  If a future version of the Open Screen Protocol uses mDNS but breaks compatibility with the metadata discovery process, it should change the DNS-SD service name to a new value, indicating a new mechanism for metadata discovery.

index.bs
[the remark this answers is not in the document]
pthatcherg: If we just don't mention port number in the spec, impls can just choose an ephemeral port (above 1024).  I don't see any value in allocating a port.  Do you see one?

index.bs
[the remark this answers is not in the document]
pthatcherg: I like it.  I replaced the current text with that text.

pthatcherg:
This conversation/review has gotten complex enough that I'm having trouble tracking all of the things going on.

I think that we're close enough to being in a good state that we should merge the PR and then resolve anything unresolved with issues and or separate PRs.

What do you think?

index.bs
```diff
@@ -235,119 +238,118 @@ Non-Functional Requirements {#requirements-non-functional}
...
+    is not yet authenticated, the requesting agent should make UI affordance for
+    indicating to the user that the display name is not yet verified.  If the
+    responding agent changes its display name, the requesting agent should
+    make UI affordance for indicating to the user that the display name has
+    changed.
 
-- model-name: The model name of the device (if it is a device), used mainly for
-  debugging purposes, but may be displayed to the user of the requesting
-  endpoint.
+-   model-name: (optional) If the agent is a hardward device, the model name of
```
mfoltzgoogle: typo in hardware

index.bs
```diff
@@ -235,119 +238,118 @@ Non-Functional Requirements {#requirements-non-functional}
...
-code of 1.  In order to keep a QUIC connection alive, an endpoint may send an
-endpoint-status-request message, and any endpoint that receives an
-endpoint-status-request message should send an endpoint-status-response message.
-Such messages should be sent more frequently than the QUIC idle_timeout
-transport parameter (see section 18 of
-[QUIC](https://tools.ietf.org/html/draft-ietf-quic-transport-16)) and QUIC PING
+Listening agents act as QUIC clients.  Advertising agents act as QUIC servers.
+
+If a listening agent wishes to receive messages from an advertising agent or an
+advertising agent wishes to send message to a listening agent, it may wish to
```
mfoltzgoogle: s/message/messages/

index.bs
mfoltzgoogle: SGTM.

index.bs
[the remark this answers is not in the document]
mfoltzgoogle: I don't have the context at hand to weigh pros and cons.  That's why I suggested opening a GitHub issue and leaving this PR as-is.

index.bs
mfoltzgoogle: Well we need to reserve ranges for those too to prevent conflicts.  I suggest proposing something in a follwup PR.

index.bs
pthatcherg: Fixed

index.bs
pthatcherg: done

PR #120, opened 2018-11-10
Add messages and protocol for Presentation API

pthatcherg:


index.bs
```diff
@@ -256,8 +256,189 @@ exchanging control messages, results, and events.
...
+messageType is embedded in the encoded CBOR type and does not need an additional
+value in the message.
+
+
+When section 6.5.6 says "Send a termination request for the presentation to its
+receiving user agent using an implementation specific mechanism", the
+controlling endpoint may send a presentation-termination-request message with
+the following values:
+
+- presentaton-id: The ID of the presentation to terminate.
```
tidoust:
```suggestion
- presentation-id: The ID of the presentation to terminate.
```

index.bs
```diff
@@ -256,8 +256,189 @@ exchanging control messages, results, and events.
...
+When a receiving endpoint receives a presentation-termination-request, it should
+send back a presentation-termination-response message to the requesting
+endpoint.  It may also notify other controlling endpoints about the termination
+by sending a presentation-termination-event message.  And it can send the same
+message if it terminates a presentation without a request from a controller to
+do so. When section section 6.5.7 says "send a termination confirmation for P
+using an implementation specific mechanism", this is the message it should send.
+This message contains the following values:
+
+- presentaton-id: The ID of the presentation that was terminated.
```
tidoust:
```suggestion
- presentation-id: The ID of the presentation that was terminated.
```

index.bs
```diff
@@ -256,8 +256,189 @@ exchanging control messages, results, and events.
...
+
+- reason: The reason the presentation was terminated.
+
+<!-- TODO: Split up reason into reason and wether it was triggered by the user or not? -->
```
tidoust:
```suggestion
<!-- TODO: Split up reason into reason and whether it was triggered by the user or not? -->
```

index.bs
```diff
@@ -256,8 +256,189 @@ exchanging control messages, results, and events.
...
+
+
+When section 6.7.1 says "it MUST listen to and accept incoming connection
+requests from a controlling browsing context using an implementation specific
+mechanism", the user agent, the receiving endpoint, must receive the and process the
```
tidoust:
```suggestion
mechanism", the user agent, the receiving endpoint, must receive and process the
```

index.bs
```diff
@@ -256,8 +256,189 @@ exchanging control messages, results, and events.
...
+presentation-connection-open-request message which contains the following
+values: 
+
+- presentaton-id: The ID of the presentation to connect to.
```
tidoust:
```suggestion
- presentation-id: The ID of the presentation to connect to.
```

index.bs
```diff
@@ -256,8 +256,189 @@ exchanging control messages, results, and events.
...
+
+
+A controlling endpoint may terminate a connection without terminating the
+presentation by sending a presentation-connection-close-request message with the
+following values:
+
+- connection-id: The ID of the connection to close.
+
+The receiving endpoint should, upon receipt of a
+presentation-connection-close-request, should send back a
```
tidoust:
```suggestion
presentation-connection-close-request, send back a
```

index.bs
```diff
@@ -256,8 +256,189 @@ exchanging control messages, results, and events.
...
+[[PRESENTATION-API|Presentation API]].  When sections are quoted here, they are
+from the [[PRESENTATION-API|Presentation API]].  When the
+[[PRESENTATION-API|Presentation API]] refers to "implementation specific
+mechanisms", this spec defines the behavior when the choosen mechanism is the
+Open Sceen Protocol.  For all references messages, see Appendix A for the full
+CDDL definitions.
+
+When section 6.4.2 says "This list of presentation displays ... is populated
+based on an implementation specific discovery mechanism", the user agent, or
+controlling endpoint, may use the the mDNS, QUIC, and endpoint-info-request
```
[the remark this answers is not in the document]
pthatcherg: This is an interesting question.  This spec is trying to do two different things at the same time:

1. Define OSP, for use by any endpoints, including non-browser endpoints.

2. Define how browser should use OSP when implementing web APIs.

When doing #2, it does seem to make more sense to use the language of the APIs.  When doing #1, it might make more sense to use non-browser language.  When doing both at the same time, which language is better?  

One option might be to break it into two sections: one that's just #1 and one that's just #2, and then use different language in each section (as if we were writing two specs).

pthatcherg:
I updated the terminology to use what I noticed was already in the "terminology" section at the start of the doc, which defines "controller" and "receiver" according to the terms in the API docs.  So, I just used those shorter names everywhere.

I also added proper links to the API docs.

But I haven't yet learned how to do internal links in a markdown/bikeshed file (use HTML <a> tags?).  If you want to explain to me how to do that, I can add some internal links.  Of course, if you're willing to do that later, I'd be happy to let you do it :).

index.bs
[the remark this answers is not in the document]
tidoust: +1 to the "two sections" option, as I suppose we want to define 1) in isolation, without referencing the Presentation API, with 2) being a mapping between API algorithm steps and relevant hooks in the protocol.

Splitting into different sections (or different specs in the end) might ease testing efforts as well, as I suppose we'll want to create a test suite at the API level, and a distinct test suite at the protocol level.

I don't know whether that split should be done right away, or if it's easier to proceed with one section for now, and refactor later on. Up to you :)

tidoust:
> But I haven't yet learned how to do internal links in a markdown/bikeshed file (use HTML tags?). If you want to explain to me how to do that, I can add some internal links. Of course, if you're willing to do that later, I'd be happy to let you do it :).

Relevant documentation is in the [Definitions](https://tabatkins.github.io/bikeshed/#definitions) and [Autolinking](https://tabatkins.github.io/bikeshed/#autolinking) sections of Bikeshed's documentation. Essentially, you'll need to use `<dfn>` tags to define terms, and use `[=toto=]` to reference them from other parts of the spec. That should be enough in most cases.

Bikeshed does not know anything about CDDL, so it won't be able to auto-dfn relevant terms in CDDL definitions. It won't be able to do syntax highlighting on CDDL blocks either, actually ([CDDL does not seem to be supported](http://pygments.org/docs/lexers/) by Pygments, used under the hoods).

Again, I'm happy to handle these editorial updates afterwards, so that you can focus on the important bits, meaning the actual definition of the protocol and mapping!

pthatcherg:
OK, I split into 2 sections.  

I'll take you up on your offer to add internal links.  You can do so later, or you could make changes directly to my PR.

index.bs
```diff
@@ -253,18 +253,251 @@ Control Protocols {#control}
...
 exchanging control messages, results, and events.
 
-Presentation API Protocol {#presentation-api}
+
+Presentation Protocol {#presentation-protocol}
 ---------------------------------------------
 
-Issue(55): Incorporate material from the
-[Protocol](protocol.md) document.
+This section defines the use of the Open Screen Protocol for staring,
```
```suggestion
This section defines the use of the Open Screen Protocol for starting,
```

index.bs
```diff
@@ -253,18 +253,251 @@ Control Protocols {#control}
...
+stopping, and controlling presenetations as defined by
+[[PRESENTATION-API|Presentation API]].  A subsequent section will
+define how APIs in maps to the protocol messages defined in this
```
btolsch: Is this sentence missing a word?

index.bs
```diff
@@ -253,18 +253,251 @@ Control Protocols {#control}
...
+
+- result: a code indicating success or failure, and the reason for the failure
+
+- connection-id: An ID that both endpoints can use to send connection messages
+to each other.  It is chosen by the receiver for ease of implementation (if the
+message receiver chooses the connection-id, it may keep the ID unique across
+connections, thus making message demuxing/routing easier).
+
+
+A controller may terminate a connection without terminating the presentation by
```
btolsch: I've been wondering if, for both close and terminate, we should exclusively use an event.  Is there any reason for the controller to wait for a close-response/termination-response?  Is there anything it should do with an error in one of those responses?

index.bs
```diff
@@ -253,18 +253,251 @@ Control Protocols {#control}
...
+values:
+
+- connection-id: The ID of the connection to close.
+
+
+The receiver should, upon receipt of a presentation-connection-close-request,
+send back a presentation-connection-close-response message with the following
+values:
+
+- result: If the close succeed or failed, and if it failed why it failed.
```
```suggestion
- result: If the close succeeded or failed, and if it failed why it failed.
```

index.bs
```diff
@@ -253,18 +253,251 @@ Control Protocols {#control}
...
+
+
+Presentation API {#presentation-api}
+---------------------------------------------
+
+This section defines how [[PRESENTATION-API|Presentation API]] uses
+the messages defined in the previous sections.  
+Non-browser endpoints can also send and receive the same messages defined here.
+If so, a non-browser endpoint must follow the same restrictions for the
+presentation-id as the a does, as defined by [[PRESENTATION-API|Presentation
```
btolsch: "as a browser would"?

index.bs
```diff
@@ -279,5 +512,145 @@ Threat model {#security-threat}
...
+  3: int ; watch-id
+}
+
+; type key 15
+presentation-url-availability-response = {
+  response
+  1: [* url-availability] ; url-availabilities
+}
+
+; type key 103
```
btolsch: Why the jump in key numbers?

index.bs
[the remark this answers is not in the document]
pthatcherg: I've split it into two sections.

index.bs
pthatcherg: Yes, added now.

index.bs
pthatcherg: The receiver could refuse to terminate for some reason (maybe it decides the controller isn't allowed to terminate that presentation).  The controller can't assume it happened until it receives the response.  The presentation API doesn't allow for a close() or terminate() method to fail asynchronously, though.  So if we wanted to allow the JS to do something, like indicate to the user that close/terminate failed, it wouldn't have a way of knowing.

index.bs
pthatcherg: This paragraph is gone now.

index.bs
pthatcherg: I added a comment explaining (Lower type keys are used for messages we expect to be more frequent and higher type keys are used for less frequently used messages)

chrisn:
A minor note, I wanted to read the changes as rendered HTML, but converting the bikeshed to HTML produced a few errors relating to indentation. I'd be happy to fix these up, but don't want to introduce whitespace only changes. LMK if you'd like me to go ahead.

pthatcherg:
Sorry, I didn't bother to run bikeshed yet.  I probably should.  It would great if you wanted to fix that.  Please, go ahead!

index.bs
```diff
@@ -253,18 +253,251 @@ Control Protocols {#control}
...
+This section defines the use of the Open Screen Protocol for starting,
+stopping, and controlling presenetations as defined by
+[[PRESENTATION-API|Presentation API]].  A subsequent section will
+define how APIs in [[PRESENTATION-API|Presentation API]] map to the
+protocol messages defined in this section.
+
+For all messages defined in this section, see Appendix A for the full
+CDDL definitions.
+
+<!-- TODO: Add a capability that indicates support for the
```
mfoltzgoogle: Can you file this as a GitHub issue, to define a capabilities structure? Or I can do a sweep later to file TODOs in the spec.

index.bs
```diff
@@ -253,18 +253,251 @@ Control Protocols {#control}
...
+presentation protocol -->
+
+To learn which receivers are [=available presentation display=]s for a
+particular URL or set of URLs, the controller may send a
+presentation-url-availability-request message with the following values:
+
+- urls: A list of presentation URLs
+
+- watch-duration: The period of time that the controller is interested in
```
mfoltzgoogle: Units?

index.bs
```diff
@@ -253,18 +253,251 @@ Control Protocols {#control}
...
+receiving updates about the URLs, should the availability change.
+
+- watch-id: An identifier the receiver may use when sending updates about URL
```
mfoltzgoogle: Legal values - a positive integer?

index.bs
```diff
@@ -253,18 +253,251 @@ Control Protocols {#control}
...
+
+<!-- TODO: Add optional HTTP response code to the response? -->
+
+To send a presentation message, the controller or receiver may send a
+presentation-connection-message with the following values:
+
+- connection-id: The ID from the presentation-start-response or
+presentation-connection-open-response messages.
+
+- message: the presentation message data.
```
mfoltzgoogle: We should define how different JS types that are accepted by PresentationConnection.send() map onto CBOR types for this data.

index.bs
```diff
@@ -253,18 +253,251 @@ Control Protocols {#control}
...
+endpoint.  It may also notify other controllers about the termination by sending
+a presentation-termination-event message.  And it can send the same message if
+it terminates a presentation without a request from a controller to do so. This
+message contains the following values:
+
+- presentation-id: The ID of the presentation that was terminated.
+
+- reason: The reason the presentation was terminated.
+
+<!-- TODO: Split up reason into reason and whether it was triggered by the user
```
mfoltzgoogle: Not required by the API spec, if I remember, but probably fine to include that level of detail in the protocol.

index.bs
mfoltzgoogle: Yeah.  close/terminate always creates a state change on the controller side.  Since the controller can't assume it can communicate with the presentation (or the device it's on) after these are called, there's no way to communicate the outcome back to the JS.

index.bs
```diff
@@ -253,18 +253,251 @@ Control Protocols {#control}
...
+
+- connection-id: The ID of the connection that was closed
+
+- reason: The reason the connection was closed
+
+- error-message: A debug message suitable for a log or perhaps presented to the
+user with more explanation as to why it was closed.
+
+<!-- TODO: Why does the Presentation API spec not mention the use of the close
+message? -->
```
mfoltzgoogle: If I remember, it was added to communicate error details when the communication channel fails, in a way that might help applications recover if they reconnect in the future.

See the notes in https://w3c.github.io/presentation-api/#sending-a-message-through-presentationconnection

In retrospect, this probably isn't and won't get used for anything in the real world, and removing closeMessage would be a reasonable spec improvement.

index.bs
```diff
@@ -253,18 +253,251 @@ Control Protocols {#control}
...
+- url-availabilities: A list of URL availability states (available, unavailable,
+or invalid).  Each state must correspond to the matching URL from the request by
+list index.
+
+
+The receivers should later (up to the current time plus request
+watch-duration) send presentation-url-availability-event messages if
+URL availabilities change.  Such events contain the following values:
+
+- watch-id: The watch-id given in the presentation-url-availability-response,
```
mfoltzgoogle: If the same controller registers multiple watches, it would be more efficient to combine the URL lists into one event when availability changes.

But somewhat more complicated to implement, and possibly not much savings.  A browser could combine its URL lists internally, before sending them off to a receiver.

index.bs
```diff
@@ -253,18 +253,251 @@ Control Protocols {#control}
...
+used to refer to the presentation URLs whose availability has changed.
+
+- url-availabilities: A list of URL availability states (available, unavailable,
+or invalid).  Each state must correspond to the URLs from the request referred
+to by the watch-id.
+
+Note that these messages are not broadcasted to all controllers. They are sent
+individually to controllers that have requested availability for the URLs that
+have changed in availability state within the watch duration of the original
+availability request.
```
mfoltzgoogle: On the other hand, it may be useful for a controller to be able to update a previous watch (URL list). Otherwise, each time a new URL availability is required, it will have to register a new watch  including all previous URLs, and get duplicate updates for them.

index.bs
```diff
@@ -253,18 +253,251 @@ Control Protocols {#control}
...
+header should be provided.
+
+The presentation ID must follow the restrictions defined by
+[[PRESENTATION-API|Presentation API]] section 6.1, in that it must
+consist of at least 16 ASCII characters.
+
+
+When the receiver receives the presentation-start-request, it should send back a
+presentation-start-response message after either the presentation URL has been
+feteched and loaded or the receiver has failed to do so. If it has failed, it
```
mfoltzgoogle:
```suggestion
feteched and loaded, or the receiver has failed to do so. If it has failed, it
```

index.bs
```diff
@@ -253,18 +253,251 @@ Control Protocols {#control}
...
+must respond with the appropriate result (such as invalid-url or timeout).  If
+it has succeeded, it must reply with a success result.  Additionally, the
+response must include the following:
+
+- connection-id: An ID that both endpoints can use to send connection messages
+to each other.  It is chosen by the receiver for ease of implementation (if the
```
mfoltzgoogle:
```suggestion
to each other.  It is chosen by the receiver for ease of implementation: if the
```

index.bs
```diff
@@ -253,18 +253,251 @@ Control Protocols {#control}
...
+message receiver chooses the connection-id, it may keep the ID unique across
+connections, thus making message demuxing/routing easier).
```
mfoltzgoogle:
```suggestion
connections, thus making message demuxing/routing easier.
```

index.bs
```diff
@@ -253,18 +253,251 @@ Control Protocols {#control}
...
+presentation-termination-request message with the following values:
+
+- presentation-id: The ID of the presentation to terminate.
+
+- reason: The reason the presentation is being terminated.
+
+
+When a receiving endpoint receives a presentation-termination-request, it should
+send back a presentation-termination-response message to the requesting
+endpoint.  It may also notify other controllers about the termination by sending
```
mfoltzgoogle: I think this is a "must" based on the spec?

https://w3c.github.io/presentation-api/#terminating-a-presentation-in-a-receiving-browsing-context

index.bs
```diff
@@ -253,18 +253,251 @@ Control Protocols {#control}
...
+
+<!-- TODO: Split up reason into reason and whether it was triggered by the user
+or not? -->
+
+
+To accept incoming connections requests from controller, a receiver
+must receive and process the presentation-connection-open-request
+message which contains the following values: 
+
+- presentation-id: The ID of the presentation to connect to.
```
mfoltzgoogle: You'll need to match the URL too.

https://w3c.github.io/presentation-api/#reconnecting-to-a-presentation

index.bs
```diff
@@ -253,18 +253,251 @@ Control Protocols {#control}
...
+
+When [[PRESENTATION-API|Presentation API]] [section
+6.3.4](https://www.w3.org/TR/presentation-api/#starting-a-presentation-connection)
+says "Using an implementation specific mechanism, tell U to create a
+receiving browsing context with D, presentationUrl, and I as
+parameters.", U (the [=controlling user agent=]) may send a
+presentation-start-request message to D (the receiver), with I for the
+presentation identifier and presentationUrl for the selected
+presentation URL.
+
```
mfoltzgoogle: Adding a section for reconnect would be good.  The spec describes reconnecting to presentations within the same browser, but leaves undefined reconnecting to presentations that are running outside the browser.  I can add a note to the spec that reconnecting to presentations on other displays is an implementation specific detail too :-)

index.bs
```diff
@@ -279,5 +512,149 @@ Threat model {#security-threat}
...
 
+Appending A: Messages
+=====================
+
+Lower type keys are used for messages we expect to be more frequent
+and higher type keys are used for less frequently used messages.  This
+is because on the wire lower values encode smaller.
+
+<pre>
+; type key 14
```
mfoltzgoogle: We could probably hack up the IDL formatter in bikeshed to pretty print these in the spec.

index.bs
```diff
@@ -279,5 +512,149 @@ Threat model {#security-threat}
...
+presentation-url-availability-request = {
+  request
+  1: [* text] ; urls
```
mfoltzgoogle: Nit: is an empty list valid here?  Similar question for presentation-url-availability-event

index.bs
```diff
@@ -279,5 +512,149 @@ Threat model {#security-threat}
...
+
+
+
+
+; type key 104
+presentation-start-request = {
+  request
+  1: text ; presentation-id
+  2: text ; url
+  3: text ; headers
```
mfoltzgoogle: [* text] or are these newline delimited?

index.bs
pthatcherg: OK

https://github.com/webscreens/openscreenprotocol/issues/123

index.bs
pthatcherg: It's in the CDDL (microseconds)

index.bs
pthatcherg: It's in the CDDL (signed int)

index.bs
pthatcherg: That's exactly what it is.  The CDDL allows bytes or text.

index.bs
pthatcherg: It was something you asked for :).

index.bs
pthatcherg: They are new-line delimited.  Basically, the straight bytes that should be written into the TCP/TLS socket.

pthatcherg: Of course, that doesn't really work for HTTP/3....  do we require impls to parse that or do we change to a list of (key, value)?

index.bs
pthatcherg: Sure, but the response would just be an empty list.

For the event, it would tell you nothing is available any more.

index.bs
pthatcherg: Do you want a TODO for that?

index.bs
pthatcherg: It might be difficult if the controller disconnected from the receiver (idle timeout and all that).  Perhaps I could make it a "should".

index.bs
pthatcherg: Done

index.bs
pthatcherg: Ok, changed to "should"

index.bs
pthatcherg: If you add text that I can quote, then I'll quote it and link to it.  

I'll put a TODO for now

index.bs
pthatcherg: It can just not renew watches and keep relatively short durations for them.

index.bs
pthatcherg: Couldn't we change the close() method to return a Promise to resolve when the close is acked?

index.bs
pthatcherg: OK, I changed to key/value pairs.

pthatcherg:
This conversation/review has gotten complex enough that I'm having trouble tracking all of the things going on.

I think that we're close enough to being in a good state that we should merge the PR and then resolve anything unresolved with issues and or separate PRs.

What do you think?

index.bs
mfoltzgoogle: Sure.

index.bs
mfoltzgoogle: Why would a controller ever want to send an empty list?

Is the empty list behavior for the event documented?

index.bs
pthatcherg: I added a TODO

index.bs
pthatcherg: It wouldn't want to send an empty list.  There's no reason.  The behavior would have to be to return a response with an empty list.  

I just added a "must not be empty" to make it clear this must never happen.

Issue #122, opened 2018-12-05
[Auth] Challenge/Response over TLS


nigelcearnshaw:
Just reviewing the outcome of the [last F2F meeting](https://www.w3.org/2018/10/26-webscreens-minutes.html#x06) during which it was resolved to further investigate the viability of a passcode related non-pake challenge-response mechanism over an already established TLS secure authenticated channel.

Am I right in assuming this requires all displays to be fitted with a confidential private key and corresponding public key certified by a certificate authority commonly trusted by most browsers?
If so, is this typically the case in the current market place?

pthatcherg:
We can discuss this more when I write up the specifics in a PR, but we would not be requiring a certificate signed by a CA.  It will likely require a self-signed certificate, which is a very common technique in p2p systems (for example, all WebRTC endpoints).

mfoltzgoogle:
PR #212 describes the certificates to be used by agents for TLS.  If there are further details needed to specify them, let's track that in a more specific issue.

PR #126, opened 2018-12-15
Add challenge/response mechanism for authentication

pthatcherg:


mfoltzgoogle:
Is this addressing Issue #13?  If we are not going to spec J-PAKE, then we should update that issue accordingly with the current plan.

Also note Issue #122

index.bs
```diff
@@ -375,7 +375,81 @@ request they are associated with.
 Authentication {#authentication}
 ================================
 
-Issue(102): Incorporate material from the \[J-PAKE](j-pake.md) document.
+In order for one agent (the challenger) to authenticate another (the responder),
+the challenger may send an authentication-request message and expect an
+authentication-response message to best sent back from the responder.  To
+mutually authenticate, this mechanism is used twice, once by each side acting as
+the challenger.  This mechanism assumes the agents share a low-entry secret,
```
mfoltzgoogle: s/entry/entropy/ ?

index.bs
```diff
@@ -375,7 +375,81 @@ request they are associated with.
 Authentication {#authentication}
 ================================
 
-Issue(102): Incorporate material from the \[J-PAKE](j-pake.md) document.
+In order for one agent (the challenger) to authenticate another (the responder),
+the challenger may send an authentication-request message and expect an
+authentication-response message to best sent back from the responder.  To
```
mfoltzgoogle: s/best/be/ ?

index.bs
```diff
@@ -375,7 +375,81 @@ request they are associated with.
...
+For all messages and objects defined in this section, see Appendix A for the full
+CDDL definitions.
+
+The challenger sends an authentication-request message with the following values:
+
+-   mechanism: The authentication mechanism being used.  This standard only
+    defines the mechanism hkdf-of-scrypt-of-psk but this field gives a place for
+    other mechanisms to be specified.
+
+-   salt: 32 random bytes (which need not be kept secret).
```
mfoltzgoogle: Define suitable method for generating these bytes?  Presumably the [TLS 1.3](https://tools.ietf.org/html/rfc8446#appendix-C.1) suggestions are appropriate here.

index.bs
```diff
@@ -375,7 +375,81 @@ request they are associated with.
...
+    greater than or equal to 14 (to avoid being too weak) and less than or equal
+    to 128 (the limit defined by scrypt).  A value of 15 is recommended (an
+    scrypt N of 2^15 or 32768).
+
+The responder replies with an authentication-response message with the following values:
+
+-   result:  If the responder was able to calculate proof of possession of the
+    shared secret, and if it failed, why it failed.
+
+-   proof: The result of running
```
mfoltzgoogle: IMO this would be shorter, and more precise, using equations to explain inputs and outputs instead of prose.  However this is a bit of a style choice.  Maybe we can add a short prose explanation here and a more explicit list of steps/equations in an appendix.

index.bs
```diff
@@ -375,7 +375,81 @@ request they are associated with.
...
+
+The certificate-fingerprint-pair object used as application-specific info above
+is serialized as CBOR (without the type or length prefix of messages in a QUIC
+stream) and contains the following values:
+
+-   challenger-fingerprint: The certificate fingerprint of the
+    challenger, used in the QUIC crypto handshake during connection
+    establishment, as defined by [RFC 8122 section
+    5](https://tools.ietf.org/html/rfc8122#section-5).  It is the output using
+    sha-256 as the hash wihtout any hexadecimal encoding or prefixing.
```
mfoltzgoogle: I didn't quite follow this.  RFC 8122 mandates upper-case hex encoding of the fingerprint.

index.bs
```diff
@@ -375,7 +375,81 @@ request they are associated with.
...
+Note: the values of 32 above (for salt length, scrypt output key length, and
+HKDF output key material length) are based on the output size of sha-256.  If a
+different hash mechanism is used in the future, these values should be updated as
+well.
+
+The certificate-fingerprint-pair object used as application-specific info above
+is serialized as CBOR (without the type or length prefix of messages in a QUIC
+stream) and contains the following values:
+
+-   challenger-fingerprint: The certificate fingerprint of the
```
mfoltzgoogle: I think the source of the certificates needs a bit more detail.  RFC 8122 describes the generation of a fingerprint but not where the certificates come from.   I assume they are from TLS 1.3 messages exchanged during QUIC crypto (Certificate/CertificateRequest)?

index.bs
```diff
@@ -375,7 +375,81 @@ request they are associated with.
...
+calculation of the proof and compares the result. If the results are the same,
+the challenger considers the responder authenticated, and considers it
+unauthenticated otherwise.  In either case, the challenger sends the responder an
+authentication-result message with the following values:
+
+-   result:  If the challenger was able to authenticate the responder or not,
+    and if not, why not.
+
+The challenger must limit the time the responder has to send a response to 60
+seconds (to avoid the possibility of brute-force attacks.)
```
mfoltzgoogle: It's probably a good idea to rate limit the challenger, since it may be possible to force a challenge to be sent through the drive-by web.   However, to prevent brute force attacks on the responder, the responder should also be rate limited.

index.bs
```diff
@@ -375,7 +375,81 @@ request they are associated with.
...
+    sha-256 as the hash wihtout any hexadecimal encoding or prefixing. 
+
+To verify that the responder's proof is correct, the challenger makes the same
+calculation of the proof and compares the result. If the results are the same,
+the challenger considers the responder authenticated, and considers it
+unauthenticated otherwise.  In either case, the challenger sends the responder an
+authentication-result message with the following values:
+
+-   result:  If the challenger was able to authenticate the responder or not,
+    and if not, why not.
```
mfoltzgoogle: If the challenger is able to authenticate the responder, is there a design for how it will be able to reconnect and authenticate?  Will it challenge again with the shared secret, or is it possible to use the PSK mechanism in TLS 1.3?

mfoltzgoogle:
Cool!  More details to work on in future PRs:
- TLS 1.3 parameters and supported extensions.
- What kinds of certificates we allow to be exchanged.

index.bs
pthatcherg: fixed

index.bs
pthatcherg: fixed

index.bs
pthatcherg: fixed

index.bs
pthatcherg: Sorry, not fixed.  Wrong comment

pthatcherg: It's for HKDF, not TLS, so the properties and details are at https://tools.ietf.org/html/rfc5869#section-3.1.  But that's a good point that we should point to that section, so I added a link to it.

index.bs
pthatcherg: Yeah, the text is a bit awkward.  I try the typical W3C style of "run the following steps".

pthatcherg: OK, I gave it a shot.  See if you like it better now.

index.bs
pthatcherg: RFC 8122 defines two things: the "certificate fingerprint and the "fingerprint attribute" (or just "fingerprint" for short).  There's no reason for us use the hash prefix or hex encoding (like the fingerprint attribute).  We just want hash of the DER representation.  I suppose we could just write that.   I'll give that a shot.

pthatcherg: OK, see if you like that better.

index.bs
pthatcherg: 8122 Section 6.2 describes the certificate presentation.  That's how you learn the remote side's certificate.  You obviously know your own.   Would you like me to reference that section?

index.bs
pthatcherg: Once authenticated, the challenger should store the certificate of the responder and allow it to connect authenticated without needing to repeat these steps and without needing TLS PSK.  But I forgot to add a paragraph explaining this, so I added on just now.

index.bs
pthatcherg: Why rate-limit a challenger?  The challenger isn't the one authenticating; the responder is.  There would be no reason to challenge many times, unless there's some way to learn the pre-shared secret using a variety of salts.  Is that what you're worried about?

The responder is rate limited here already in two ways: 1 by scrypt.  It can only compute so many responses.   2 by the challenger which will only issues so many challenges (probably one per QUIC connection).  

Or perhaps that's what you meant by rate limit the challenger?  Only attempt N (perhaps = 1) times per QUIC connection?

index.bs
btolsch: I'd also assume the certificates are supposed to come from the TLS handshake messages.  However, FWIW, 8122-6.2 is confusing to me since it's mostly discussing an SDP context.

index.bs
mfoltzgoogle: If an application acting as a challenger can force a challenge to be sent to a specific responder, it can create a denial of service to the responder.   Most reasonable implementations should mitigate this by aggregating connections across applications/sites, so an application can't force multiple connections to be attempted to the same receiver.  However, rate limiting in the challenger implementation would be an added layer of protection.

Since I wasn't thinking of crypto vulnerabilities, it feels more like  a security mitigation in the implementation, instead of a spec issue for the authentication handshake.  In which case I'm happy to discuss rate limiting in my other pull request - no changes required here.

mfoltzgoogle:
Filed Issue #135 to discuss TLS certificate types and extensions.  This looks good to merge from me.

Issue #123, opened 2018-12-06
Define a capabilities mechanism


pthatcherg:
We need a way to know which parts of the protocol an endpoint supports: does it allow Presentations? Remote Playback?  Streaming?  And if we make a new version of a protocol, does it support that?  We need a mechanism for this.

mfoltzgoogle:
Should ensure that any capabilities disclosed before authentication are discussed in the Security & Privacy section, and the possible implications thereto.

rektide:
My ask for audio/video capabilities appears to be the primary thing implemented in the capabilities PR, https://github.com/webscreens/openscreenprotocol/pull/140/files . 👍

mfoltzgoogle:
Closing this as we've landed spec changes:
- For agents to exchange capabilities
- A list of capabilities defined by the spec
- A way for agents to advertise extended (custom) capabilities
- A registry of extended capabilities
- Noting that capabilities are advertised to any network client and are public information.

The remaining open issue around capabilities is #200, but that is specific to receives-audio and receives-video.   Closing this issue for now based on the current state of the spec.

PR #133, opened 2019-01-03
Adds security and privacy section to the spec.

mfoltzgoogle:
Addresses Issue #13: Propose security mechanisms for the OSP

This is a draft version of the Security and Privacy section of the Open Screen Protocol spec, following the W3C's Security and Privacy Questionnaire [1].  It addresses the following:

* Threats
* Considerations specific to the OSP
* Considerations from the Security and Privacy sections of the Presentation API and Remote Playback API
* Some possible mitigations, with issues filed to investigate some specific areas further

Would appreciate wide review.  In particular, if there are topics from previous discussions missing, can add them here or in a followup PR.

[1] https://www.w3.org/TR/security-privacy-questionnaire/#questions

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -376,7 +376,81 @@ request they are associated with.
 Authentication {#authentication}
 ================================
 
-Issue(102): Incorporate material from the \[J-PAKE](j-pake.md) document.
+In order for one agent (the challenger) to authenticate another (the responder),
+the challenger may send an authentication-request message and expect an
+authentication-response message to best sent back from the responder.  To
+mutually authenticate, this mechanism is used twice, once by each side acting as
+the challenger.  This mechanism assumes the agents share a low-entry secret,
+such as a number or a short password that could be entered by a user on a
+keyboard or TV remote control.
+
+For all messages and objects defined in this section, see Appendix A for the full
+CDDL definitions.
+
+The challenger sends an authentication-request message with the following values:
+
+-   mechanism: The authentication mechanism being used.  This standard only
+    defines the mechanism hkdf-of-scrypt-of-psk but this field gives a place for
+    other mechanisms to be specified.
+
+-   salt: 32 random bytes (which need not be kept secret). 
+
+-   cost: log base 2 of the cost parameter (N) for scrypt defined in [RFC
+    7914 section 2](https://tools.ietf.org/html/rfc7914#section-2).  It must be
+    greater than or equal to 14 (to avoid being too weak) and less than or equal
+    to 128 (the limit defined by scrypt).  A value of 15 is recommended (an
+    scrypt N of 2^15 or 32768).
+
+The responder replies with an authentication-response message with the following values:
+
+-   result:  If the responder was able to calculate proof of possession of the
+    shared secret, and if it failed, why it failed.
+
+-   proof: The result of running
+    [HKDF](https://tools.ietf.org/html/rfc5869) on the result of running
+    [scrypt](https://tools.ietf.org/html/rfc7914) on the shared secret.  For
+    scrypt, the cost parameter (N) used is the 2 to the power of the cost from
+    the authentication-request message, the block size (r) is 8, the
+    parallelization parameter (p) is 1, and the key length derived is 32.  For
+    HKDF, both the extract and expand steps are used, the hash function used is
+    sha-256, the salt used is the salt from the authentication-request message,
+    the application-specific info is a serialized certificate-fingerprint-pair
+    object (defined below), and the length of the output key matrial (L) is 32.
+
+Note: the values of 32 above (for salt length, scrypt output key length, and
+HKDF output key material length) are based on the output size of sha-256.  If a
+different hash mechanism is used in the future, these values should be updated as
+well.
+
+The certificate-fingerprint-pair object used as application-specific info above
+is serialized as CBOR (without the type or length prefix of messages in a QUIC
+stream) and contains the following values:
+
+-   challenger-fingerprint: The certificate fingerprint of the
+    challenger, used in the QUIC crypto handshake during connection
+    establishment, as defined by [RFC 8122 section
+    5](https://tools.ietf.org/html/rfc8122#section-5).  It is the output using
+    sha-256 as the hash wihtout any hexadecimal encoding or prefixing.
+
+-   responder-fingerprint: The certificate fingerprint of the
+    responder, obtained from the QUIC crypto handshake during connection
+    establishment, as defined by [RFC 8122 section
+    5](https://tools.ietf.org/html/rfc8122#section-5).  It is the output using
+    sha-256 as the hash wihtout any hexadecimal encoding or prefixing. 
+
+To verify that the responder's proof is correct, the challenger makes the same
+calculation of the proof and compares the result. If the results are the same,
+the challenger considers the responder authenticated, and considers it
+unauthenticated otherwise.  In either case, the challenger sends the responder an
+authentication-result message with the following values:
+
+-   result:  If the challenger was able to authenticate the responder or not,
+    and if not, why not.
+
+The challenger must limit the time the responder has to send a response to 60
+seconds (to avoid the possibility of brute-force attacks.)
+
+
 
 Control Protocols {#control-protocols}
 ============================
@@ -510,8 +584,7 @@ message which contains the following values:
 
 -   presentation-id: The ID of the presentation to connect to.
 
-- url: The URL of the presentation to connect to.
-
+-   url: The URL of the presentation to connect to. 
 
 The receiver should, upon receipt of a
 presentation-connection-open-request message, send back a
@@ -966,6 +1039,53 @@ response = (
 )
 
 request-id = uint
+
+; type key 1001 
+authentication-request = {
+  request
+  1: authentication-mechanism ; mechanism
+  2: bytes ; salt
+  3: uint ; cost
+}
+
+; type key 1002 
+authentication-response = {
+  response
+  1: authentication-response-result ; result
+  2: bytes ; proof
+}
+
+certificate-fingerprint-pair = [
+  challenger-fingerprint: bytes
+  responder-fingerprint: bytes
+]
+
+; type key 1003
+authentication-result = {
+  1: authentication-result-result ; result
+}
+
+
+authentication-mechanism = &(
+  hkdf-of-scrypt-of-psk: 1
+)
+
+authentication-response-result = &(
+  ok: 0
+  unknown-error: 1
+  mechanism-unknown: 2
+  salt-too-small: 3
+  cost-too-low: 4
+  cost-too-high: 5
+  secret-unknown: 6
+  calculation-took-too-long: 7
+)
+
+authentication-result-result = &(
+  authenticated: 0
+  unknown-error: 1
+  proof-invalid: 2
+)
 
 ; type key 14
 presentation-url-availability-request = {
--- a/index.bs
+++ b/index.bs
@@ -632,18 +632,159 @@ Issue(12): Propose control protocol for Remote
 Playback API.
 
 
-Security and Privacy {#security}
-================================
+Security and Privacy {#security-privacy}
+====================
 
-The Open Screen Protocol should, at a minimum, conform to the security and
-privacy guidelines recommended by the [[PRESENTATION-API][Presentation API]] and
-the [[REMOTE-PLAYBACK][Remote Playback API]].
+The Open Screen Protocol allows two networked user agents to discover each other
+and exchange user and application data.  As such, its security and privacy
+considerations should be closely examined.  We first evaluate the protocol
+itself using the W3C [[SECURITY-PRIVACY-QUESTIONNAIRE]].  We then examine
+whether the security and privacy guidelines recommended by the
+[[PRESENTATION-API]] and the [[REMOTE-PLAYBACK] are met.
 
-In addition, the Open Screen Protocol itself adds additional security and privacy
-considerations.
+Threat Models {#threat-models}
+--------------------------------
 
-Presentation API
-----------------
+### Passive Network Attackers ### {#passive-network-attackers}
+
+The Open Screen Protocol should assume that all parties that are connected to
+the same LAN as the user agents, either through a wired connection or through
+WiFi, are able to observe all data flowing between Open Screen controllers and
+receivers.
+
+These parties will be able collect any data exposed through unencrypted
+messages, such as mDNS records and the QUIC handshake.
+
+These parties may attempt to learn cryptographic parameters by observing data
+flows on the QUIC connection, or by observing cryptographic timing.
+
+### Active Network Attackers ### {#active-network-attackers}
+
+In addition, all parties with access to the LAN will be able to manipulate data
+exchanged between controllers and receivers.  They will be able to inject
+traffic into both parties and attempt to inititate new QUIC connections.  These abilities
+can be used to attempt the following:
+
+*   Impersonate a new receiver or one already trusted by the user, in an attempt
+    to convince the user to authenticate to it.
+*   Connect to a receiver and query its capabilities.
+*   Connect to and control a presentation or remote playback, or extract data
+    from the application state of the presentation or remote playback.
+    
+One particular attack of concern is misconfigured or compromised routers that
+expose local network devices to the Internet.  This vector of attack has been
+used by remote parties to take control of printers and smart TVs by connecting
+to local network services that should normally be inaccessible from the
+Internet.
+
+### Denial of Service ### {#denial-of-service}
+
+Parties with access to the LAN may attempt to deny access to receivers by
+controllers, or vice versa.  For example, an active attacker my attempt to start
+a large number of QUIC connections on a receiver in an attempt to block
+legitimate connections or exhaust available resources on the presentation
+display.  They may also multicast spurious DNS-SD records in an attempt to
+exhaust the cache capacity for mDNS listeners.
+
+### Same-Origin Policy Violations ### {#same-origin-policy-violations}
+
+The Presentation API allows cross-origin communication between controlling pages
+and presentations with the consent of each origin (through their use of the
+API).  This is similar to cross-origin communication via {{Worker/postMessage}}.  The
+Open Screen Protocol does not convey origin information between controllers and
+receivers.
+
+The [=presentation ID=] carries some protection against unrestricted
+cross-origin access, but authentication of the parties connected by a
+{{PresentationConnection}} can be done at the application level.
+
+### Third Party Tracking ### {#third-party-tracking}
+
+The Open Screen Protocol does not give third party origins additional access to
+presentations or remote playback state.
+
+Open Screen Protocol Considerations {#security-privacy-questions}
+-----------------------------------
+
+### Personally Identifiable Information & High-Value Data ### {#personally-identifiable-information}
+
+The following data exchanged by the protocol can be personally identifiable and
+thus is high value data:
+
+1. Presentation URLs and availability results
+1. Presentation IDs
+1. Presentation connection IDs
+1. Presentation connection messages
+1. Remote playback URLs
+1. Remote playback commands and status messages
+
+Presentation display friendly names, model names, and capabilities, while not
+considered personally identifiable, is important to protect to preserve the
+integrity of the discovery process.
+
+The following data cannot be reasonably secured and should be considered public
+and untrusted data:
+
+1. IP addresses and ports used by the Open Screen Protocol.
+1. Data advertised through mDNS, including the certificate fingerprint and the 
+    metadata version.
+   
+### Cross Origin State Considerations ### {#cross-origin-state}
+
+Access to origin state across browsing sessions is possible through the
+Presentation API by reconnecting to a presentation that was started by a
+previous session. This sceanrio is addressed in
+[[PRESENTATION-API#cross-origin-access]].
+
+Presentation display availability and remote playback availability are states
+that may be available cross-origin depending on the user's network context.
+Exposure of this data to the Web is also discussed in
+[[PRESENTATION-API#personally-identifiable-information]] and
+[[REMOTE-PLAYBACK#personally-identifiable-information]].
+
+### Origin Access to Other Devices ### {#origin-access-devices}
+
+By design, the Open Screen Protocol allows access to presentation displays from
+the Web.  By implementing the protocol, these devices are knowingly making
+themselves available to the Web and should be designed accordingly.
+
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -632,19 +632,281 @@ Issue(12): Propose control protocol for Remote
...
+    number of times.
+  
+Flagging means that the user is notified, or in some cases they are required to
+re-authenticate to the presentation display to verify its identity.
+  
+The second is through management of the shared secret during mutual
+authentication:
+
+* Rotate the shared secret to prevent brute force attacks.
+* Use an increasing backoff to repond to authentication challenges, also to
```
```suggestion
* Use an increasing backoff to respond to authentication challenges, also to
```

index.bs
```diff
@@ -632,19 +632,281 @@ Issue(12): Propose control protocol for Remote
...
+    prevent brute force attacks.
+* Use a cryptographically sound source of entropy to generate the shared secret.
+* Require the end user to manually type the shared secret - shown only the
```
```suggestion
* Require the end user to manually type the shared secret - shown only on the
```

index.bs
```diff
@@ -632,19 +632,281 @@ Issue(12): Propose control protocol for Remote
...
+controllers and receivers from traffic from the broader Internet.  Open Screen
+Protocol implementations that are only intended to work on the LAN should filter
+packets from non-local IP addresses.  Implementations can also use the ARP cache
+to detect attempts to spoof local network IP addresses.
+
+Issue(131): [Security] Mitigations for remote network attackers
+
+### Denial of service ### {#denial-of-service-mitigations}
+
+It will be difficult to completely prevent denial service of attacks that
```
```suggestion
It will be difficult to completely prevent denial of service attacks that
```

index.bs
```diff
@@ -632,19 +632,281 @@ Issue(12): Propose control protocol for Remote
...
+that attempts to compromise the target device by exploiting parsing
+vulnerabilities.
+
+CBOR is intended to be less vulnerable to such attacks relative to alternatives
+like JSON and XML.  Still, implementations should be thoroughly tested using
+approaches like [fuzz testing](https://en.wikipedia.org/wiki/Fuzzing).
+
+Where possible, Open Screen Protocol implementations (including the content
+rendering components) should use defense-in-depth techniques like <a
+href="https://en.wikipedia.org/wiki/Sandbox_(computer_security)">sandboxing</a>
```
btolsch: Are markdown links or HTML <a> tags preferred for hyperlinks?  Both are used within a few lines of each other, which seems odd.

index.bs
```diff
@@ -632,19 +632,281 @@ Issue(12): Propose control protocol for Remote
...
+API).  This is similar to cross-origin communication via {{Window/postMessage()}}
+with a target origin of `*`.  However, the Presentation API does not convey
+source origin information with each message.  Therefore, the Open Screen
+Protocol does not convey origin information between controllers and receivers.
+
+The [=presentation ID=] carries some protection against unrestricted
+cross-origin access; but, rigorous authentication of the parties connected by a
+{{PresentationConnection}} must be done at the application level.
+
+Open Screen Protocol Considerations {#security-privacy-questions}
```
```suggestion
Open Screen Protocol Security and Privacy Considerations {#security-privacy-questions}
```

index.bs
```diff
@@ -632,19 +632,281 @@ Issue(12): Propose control protocol for Remote
...
+
+Issue(131): [Security] Mitigations for remote network attackers
+
+### Denial of service ### {#denial-of-service-mitigations}
+
+It will be difficult to completely prevent denial service of attacks that
+originate on the user's local area network.  Open Screen Protocol
+implementations can refuse new connections, rate limit the rate of messages from
+existing connections, or limit the number of mDNS records cached in an attempt
+to allow existing activities to continue in spite of such an attack.
```
btolsch: WDYT about limiting the number of mDNS records cached from a particular responder?

index.bs
```diff
@@ -632,19 +632,281 @@ Issue(12): Propose control protocol for Remote
...
 
-Security and Privacy {#security}
-================================
-
-Issue(13): Describe security architecture.
+Security and Privacy {#security-privacy}
+====================
 
-Data to be secured {#security-data}
+The Open Screen Protocol allows two networked user agents to discover each other
```
pthatcherg: user agents => agents

index.bs
```diff
@@ -632,19 +632,281 @@ Issue(12): Propose control protocol for Remote
...
+requirements.
+
+Threat Models {#threat-models}
+--------------------------------
+
+### Passive Network Attackers ### {#passive-network-attackers}
+
+The Open Screen Protocol should assume that all parties that are connected to
+the same LAN, either through a wired connection or through WiFi, are able to
+observe all data flowing between Open Screen controllers and receivers.
```
pthatcherg: perhaps just "between agents".  We might have more than controllers and receivers in the future.

index.bs
```diff
@@ -632,19 +632,281 @@ Issue(12): Propose control protocol for Remote
...
+
+These parties will be able collect any data exposed through unencrypted
+messages, such as mDNS records and the QUIC handshake.
```
pthatcherg: the QUIC handshake => QUIC handshakes

?

index.bs
```diff
@@ -632,19 +632,281 @@ Issue(12): Propose control protocol for Remote
...
+
+These parties may attempt to learn cryptographic parameters by observing data
+flows on the QUIC connection, or by observing cryptographic timing.
+
+### Active Network Attackers ### {#active-network-attackers}
+
+In addition, all parties connected to the LAN will be able to manipulate data
```
pthatcherg: All parties?  Or just some parties?

index.bs
```diff
@@ -632,19 +632,281 @@ Issue(12): Propose control protocol for Remote
...
+exchanged between controllers and receivers.  They will be able to inject
+traffic into existing QUIC connections and attempt to inititate new QUIC
+connections.  These abilities can be used to attempt the following:
+
+*   Impersonate a new receiver or one already trusted by the user, in an attempt
+    to convince the user to authenticate to it.
```
pthatcherg: Could impersonate a controller as well (future scenarios might have the TV be a controller).  Perhaps just say agent?

index.bs
```diff
@@ -632,19 +632,281 @@ Issue(12): Propose control protocol for Remote
...
+*   Connect to a receiver and query its capabilities.
```
pthatcherg: Are we going to do capability-query before or after authentication?  We probably want some capabilities to be known before authentication.  But perhaps not all.  I can't think of a capability that you wouldn't want known.  Can you?

index.bs
```diff
@@ -632,19 +632,281 @@ Issue(12): Propose control protocol for Remote
...
 
-Threat model {#security-threat}
+### Personally Identifiable Information & High-Value Data ### {#personally-identifiable-information}
+
+The following data exchanged by the protocol can be personally identifiable
+and/or high value data:
+
+1. Presentation URLs and availability results
+1. Presentation IDs
+1. Presentation connection IDs
```
pthatcherg: Why are connection IDs valuable?

index.bs
```diff
@@ -632,19 +632,281 @@ Issue(12): Propose control protocol for Remote
...
 -----------------------------------
 
-Threat model {#security-threat}
+### Personally Identifiable Information & High-Value Data ### {#personally-identifiable-information}
+
+The following data exchanged by the protocol can be personally identifiable
+and/or high value data:
+
+1. Presentation URLs and availability results
+1. Presentation IDs
```
pthatcherg: Should we explain why presentation IDs are valuable?    The rest seem more obvious.

index.bs
```diff
@@ -632,19 +632,281 @@ Issue(12): Propose control protocol for Remote
...
+1. Presentation connection IDs
+1. Presentation connection messages
+1. Remote playback URLs
+1. Remote playback commands and status messages
+
+Presentation display friendly names, model names, and capabilities, while not
+considered personally identifiable, is important to protect to preserve the
+integrity of the discovery and authentication process.
```
pthatcherg: What do you mean "preserve the integrity"?  Do you mean an attacker can't change them or impersonate new values of them?

index.bs
```diff
@@ -632,19 +632,281 @@ Issue(12): Propose control protocol for Remote
...
+
+The following data cannot be reasonably secured and should be considered public
+and untrusted data:
+
+1. IP addresses and ports used by the Open Screen Protocol.
+1. Data advertised through mDNS, including the certificate fingerprint and the 
+    metadata version.
```
pthatcherg: When you say "secured", it would probably be more precise to say "made confidential", since we can ensure integrity, just not confidentiality.  

We also can't make display names confidential unless we required a user to authenticate before picking the thing to authenticate to (which might be possible if we did QR code scanning type things).

index.bs
```diff
@@ -632,19 +632,281 @@ Issue(12): Propose control protocol for Remote
...
+devices are knowingly making themselves available to the Web and should be
+designed accordingly.
+
+Below, we discuss mitigation steps to prevent malicious use of these devices.
+
+### Incognito Mode ### {#incognito-mode}
+
+The Open Screen Protocol does not distinguish between the user agent's normal
+browsing and incognito modes, and should operate identically regardless of which
+mode is in use.
```
pthatcherg: Will this always be true?  It seems like a very broad statement that might not always be true.

index.bs
```diff
@@ -632,19 +632,281 @@ Issue(12): Propose control protocol for Remote
...
+
+It's recommended that controllers use separate authentication contexts and QUIC
+connections for normal and incognito profiles from the same user agent instance.
+This prevents Open Screen receivers from correlating activity between a user's
+normal profile and their incognito profile.
+
+### Persistent State ### {#persistent-state}
+
+An implementation of the Open Screen Protocol is likely to persist the identity
```
pthatcherg: We have a word for " implementation of the Open Screen Protocol": agent.  You could just use that here.

index.bs
```diff
@@ -632,19 +632,281 @@ Issue(12): Propose control protocol for Remote
...
+
+Below, we discuss mitigation steps to prevent malicious use of these devices.
+
+### Incognito Mode ### {#incognito-mode}
+
+The Open Screen Protocol does not distinguish between the user agent's normal
+browsing and incognito modes, and should operate identically regardless of which
+mode is in use.
+
+It's recommended that controllers use separate authentication contexts and QUIC
```
pthatcherg: A browser might be a receiver as well.  Should it also not do the same when being a receiver?

index.bs
```diff
@@ -632,19 +632,281 @@ Issue(12): Propose control protocol for Remote
...
-Mitigations and defenses {#security-defenses}
----------------------------------------------
+[[PRESENTATION-API#security-and-privacy-considerations]] place these
+requirements on the Open Screen Protocol:
+
+1.  Presentation URLs and presentation IDs should remain private among the
+    parties that are allowed to connect to a presentation, per the
+    cross-origin access guidelines.
+1.  Controllers and recievers should be notified when multiple connections have
+    been made to a presentation, per the user interface guidelines.
```
pthatcherg: Controllers need to be notified when _other_ controllers connect to the same presentation?  Or just the receiver needs to notify its user?

index.bs
```diff
@@ -632,19 +632,281 @@ Issue(12): Propose control protocol for Remote
...
+1.  Messaging between presentations and controllers should be authenticated and
```
pthatcherg: Do you mean between receivers and controllers?

index.bs
```diff
@@ -632,19 +632,281 @@ Issue(12): Propose control protocol for Remote
...
+
+Mitigation Strategies {#security-mitigations}
+--------------------------------------------
+
+### Local passive network attackers ### {#local-passive-mitigations}
+
+Local passive attackers may attempt to harvest data about user activities and
+device capabilties using the Open Screen Protocol.  The main strategy to address
+this is data minimization, by only exposing opaque public key fingerprints
+before user-mediated authentication takes place.
```
pthatcherg: As mentioned, we'll need some capabilities before authentication.  But we'll need to figure out what those are.

index.bs
```diff
@@ -632,19 +632,281 @@ Issue(12): Propose control protocol for Remote
...
+
+Passive attackers may also attempt timing attacks to learn cryptographic the
+cryptographic parameters of the TLS 1.3 QUIC connection.
+
```
pthatcherg: cryptographic the cryptographic => the cryptographic

index.bs
```diff
@@ -632,19 +632,281 @@ Issue(12): Propose control protocol for Remote
...
+Protocol prevents a man-in-the-middle from impersonating a controller or
+receiver, without knowledge of a shared secret.  However, it is possible for an
+attacker to impersonate an existing, trusted display or a newly discovered
+display that is not yet authenticated and try to convince the user to
+authenticate it.
+
+This can be addressed through a combination of techniques.  The first is flagging:
+
+* Flag advertised displays whose friendly name, public key fingerprint, or
+    IP:port collide with an already trusted display.
```
pthatcherg: I think we should flag when we see the friendly name with a different fingerprint, but not if we see the same IP:port with a different friendly name or FP (an AP may reuse IPs and client may choose ports fairly statically).  And I don't think we should flag the same FP but different IP:port (since IP is likely to change).  The only other combination seems to be same FP but different friendly name, which should also be flagged.

index.bs
```diff
@@ -632,19 +632,281 @@ Issue(12): Propose control protocol for Remote
...
+* Flag already-trusted displays whose metadata has changed.
```
pthatcherg: What do you mean by metadata here?  That could include a lot of things.  Do we really want to show the user something every time anything changes?

index.bs
```diff
@@ -632,19 +632,281 @@ Issue(12): Propose control protocol for Remote
...
+* Flag controllers or receivers that fail the authentication challenge a certain
+    number of times.
```
pthatcherg: Shouldn't we also notify the user of the authentication state of a particular device so it can know the difference between a new device and a previously-trusted device even when the names aren't exactly the same (perhaps the impersonation is done by using unicode characters that look just like normal characters so the strings don't match but the look the same).

index.bs
```diff
@@ -632,19 +632,281 @@ Issue(12): Propose control protocol for Remote
...
+  
+Flagging means that the user is notified, or in some cases they are required to
+re-authenticate to the presentation display to verify its identity.
```
pthatcherg: When would re-auth be needed?  When the friendly name changes?

index.bs
```diff
@@ -632,19 +632,281 @@ Issue(12): Propose control protocol for Remote
...
+  
+The second is through management of the shared secret during mutual
+authentication:
+
+* Rotate the shared secret to prevent brute force attacks.
```
pthatcherg: What shared secret?  There is no long-term shared secret.  There are only trusted certs.    There is nothing to rotate.  Although, I suppose you could un-trust certs if you wanted to.

index.bs
```diff
@@ -632,19 +632,281 @@ Issue(12): Propose control protocol for Remote
...
+
+Issue(130): [Security] Review attack and mitigation considerations for TLS 1.3
+
+### Remote active network attackers ### {#remote-active-mitigations}
+
+Unfortunately, we cannot rely on network devices to fully protect Open Screen
+controllers and receivers from traffic from the broader Internet.  Open Screen
+Protocol implementations that are only intended to work on the LAN should filter
+packets from non-local IP addresses.  Implementations can also use the ARP cache
+to detect attempts to spoof local network IP addresses.
```
pthatcherg: I think encouraging implementations to filter based on IP address seems risky because if it's traffic going through a NAT that is misconfigured, the NAT will likely remap the addresses to local one, and so this technique might give a false sense of security.

It might be better to say that all endpoints should be secure against attack from any Internet host.  That would work better as we expand to cross-network use cases and using ICE as well.

index.bs
```diff
@@ -632,19 +632,281 @@ Issue(12): Propose control protocol for Remote
...
+
+Issue(131): [Security] Mitigations for remote network attackers
+
+### Denial of service ### {#denial-of-service-mitigations}
+
+It will be difficult to completely prevent denial service of attacks that
+originate on the user's local area network.  Open Screen Protocol
+implementations can refuse new connections, rate limit the rate of messages from
+existing connections, or limit the number of mDNS records cached in an attempt
```
pthatcherg: How would you rate limit incoming messages?  Close a connection when too many messages come in?

index.bs
mfoltzgoogle: This is actually to work around a bug in Bikeshed :-)
See https://github.com/tabatkins/bikeshed/issues/1419

index.bs

index.bs

index.bs

index.bs
mfoltzgoogle: Clarified with an example (compromised router) that has visibility into all the packets flowing between agents.

index.bs
mfoltzgoogle: Done, and I went through and replaced all "controller and receiver" mentions with "agent."

index.bs
```diff
@@ -632,19 +632,281 @@ Issue(12): Propose control protocol for Remote
...
+### Origin Access to Other Devices ### {#origin-access-devices}
+
+By design, the Open Screen Protocol allows access to presentation displays and
+remote playback devices from the Web.  By implementing the protocol, these
+devices are knowingly making themselves available to the Web and should be
+designed accordingly.
+
+Below, we discuss mitigation steps to prevent malicious use of these devices.
+
+### Incognito Mode ### {#incognito-mode}
```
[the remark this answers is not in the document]
mfoltzgoogle: Agreed - I updated the last statement to apply across all profiles:

> This prevents Open Screen receivers from correlating activity among a user's
profiles (both normal and incognito).

Let me know if you think there's further clarification needed.

index.bs
mfoltzgoogle: Agreed.  Updated text here.

index.bs
mfoltzgoogle: Changed "secured" to "made confidential" and included the display name prefix (as it's part of mDNS).

index.bs
mfoltzgoogle: Implementations could go beyond the spec to propagate information that would indicate if incognito is in use (even if it's a bad idea).  Added a caveat here.

index.bs
mfoltzgoogle: Updated "implementation" to "agent" here and elsewhere.

index.bs
mfoltzgoogle: A connection id maps onto a PresentationConnection (i.e. a player/controller in a multiplayer game).  If you expose them, one connection could impersonate another.

index.bs
mfoltzgoogle: I would prefer post-authentication by default.  The capabilities expose some facts about the devices in your vicinity to third parties, which could collect/mine that data for analytics or advertising purposes.   Also, we don't want the capabilities to be modified by a MITM (see the PII section).

If there's a use case for exposing them pre-authentication, we can explore that tradeoff.

index.bs
mfoltzgoogle: The Presentation API security and privacy section talks a little bit about this, added a clarification here.

index.bs
mfoltzgoogle: Yes.  Clarified the wording here.

index.bs
mfoltzgoogle: What are they?

baylesj: I'm planning on address the capabilities in a separate issue, here: https://github.com/webscreens/openscreenprotocol/issues/123.

Can we assign that issue to me? I currently don't have permission for that.

index.bs
mfoltzgoogle: Yes it should. Updated this language to require separate authentication contexts for all Open Screen agents.

index.bs
mfoltzgoogle: Both.  The guideline is about notifying all connected endpoints that a third party is now able to interact with the presentation.

It looks like the spec only requires a presentation-connection-open-response to be sent to the endpoint that requested the connection. 

Filed https://github.com/webscreens/openscreenprotocol/issues/143

index.bs
mfoltzgoogle: Fixed

index.bs
mfoltzgoogle: @pthatcherg I updated Issue #123 to make sure that this gets updated with capability info.
@baylesj I think the assignee list has to be updated by the repository owner.

index.bs
mfoltzgoogle: Further fixes can be addressed in response to Issue #143.

index.bs
mfoltzgoogle: @baylesj Contacted @anssiko about enabling issue assignment.  Further fixes to this section can happen in the open PR on capabilities.

index.bs
mfoltzgoogle: Fixed

index.bs
mfoltzgoogle: I broke this into two criteria:
- Flag FP collision.
- Flag new friendly name for a previously trusted FP.
I think that covers the cases you describe.

index.bs
mfoltzgoogle: Right now the metadata includes the display name and the model name.  If either of those changes that should cause the device to be flagged IMO.

mfoltzgoogle: I updated this to refer to specifically the data in `agent-info.`

index.bs
mfoltzgoogle: Good point!  Added name similarity as a criterion for detecting an attempt at impersonation.

index.bs
mfoltzgoogle: This is the case when the metadata for an already-trusted agent changes.  Updated the text to clarify.

index.bs
mfoltzgoogle: This is the "low entropy secret" (number, short password) described in the Authentication section.  I assume that the same secret would normally be used for the challenge-response on both directions (i.e., one agent advertises, the other consumes), but please correct me if I'm mistaken.

Updated the text here to be clearer.

index.bs
mfoltzgoogle: Agreed.  Updated this section.

index.bs
mfoltzgoogle: Yeah ... pretty much.  Rephrased to be more specific that I'll close the connection here.

mfoltzgoogle:
@pthatcherg @btolsch I believe I've resolved all outstanding comments, and I'm going to merge now.  All should free to file any additional topics as issues with tag Security.  There are already [several open items](https://github.com/webscreens/openscreenprotocol/labels/Security) which will require further work on this section or the Authentication section of the spec.

PR #136, opened 2019-01-14
Fix indentation that was causing bikeshed errors.

markafoltz:
FYI @pthatcherg 

Also regenerates index.html.


pthatcherg:
Thanks.  I started running bikeshed myself when working on the Remote Playback API PR.  So hopefully this won't happen any more.

PR #140, opened 2019-03-08
Add capabilities 

baylesj:
This patch adds specifications for getting device capabilities through the open screen protocol. Specifically, what types of actions and media does the receiver support?

NOTE: This patch is based on @pthatcherg's remote playback pull request, so make sure to exclude those commits when diffing!

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -297,38 +297,35 @@ mechanism for metadata discovery.
 Transport and metadata discovery with QUIC {#transport}
 =======================================================
 
-If a listening agent ("listener") wants to connect to an advertising agent
-("advertiser"), or to learn further metadata about it, it initiates a [[QUIC]]
-connection to the IP and port from the SRV record.  Prior to authentication,
-a message may be exchanged (such as further metadata), but such info should be
-treated as unverified (such as indicating to a user that a display name of an
-unauthenticated agent is unverified). Note that both listeners and advertisers
-may request information about each other (performing as a "requester" to
-query a "requestee").
+If an agent wants to connect to or learn further metadata about another agent,
+it initiates a [[QUIC]] connection to the IP and port from the SRV record.
+Prior to authentication, a message may be exchanged (such as further metadata),
+but such info should be treated as unverified (such as indicating to a user that
+a display name of an unauthenticated agent is unverified).
 
-To learn further metadata, devices may send an agent-info-request
+To learn further metadata, an agent may send an agent-info-request
 message (see [[#appendix-a]]) and receive back an agent-info-response message.
-Note that both the listener and advertiser may send this request to learn about
-the capabilities of the other device.
+Any agent may send this request to learn about the capabilities of
+another device.
 
-The messages may contain the following properties:
+The agent-info-response messages contain the following properties:
 
 : display-name (required)
-:: The display name of the requestee, intended to be displayed to a user by the
+:: The display name of the agent, intended to be displayed to a user by the
      requester. The requester should indicate through the UI if the device
      is not authenticated or if the display name changes.
 
 : model-name (optional)
-:: If the requestee is a hardware device, the model name of
+:: If the agent is a hardware device, the model name of
     the device.  This is used mainly for debugging purposes, but may be
     displayed to the user of the requesting agent.
 
 : supports-audio (optional)
-:: The requestee has to indicate that it supports audio streaming. If false
+:: The agent has to indicate that it supports audio streaming. If false
     or not included, it is assumed audio streaming is not supported.
 
 : supports-video (optional)
-:: The requestee has to indicate that it supports video streaming. If false
+:: The agent has to indicate that it supports video streaming. If false
     or not included, it is assumed video streaming is not supported.
 
 
--- a/index.bs
+++ b/index.bs
@@ -308,11 +308,11 @@ message (see [[#appendix-a]]) and receive back an agent-info-response message.
 Any agent may send this request to learn about the capabilities of
 another device.
 
-The agent-info-response messages contain the following properties:
+The agent-info-response message contains the following properties:
 
 : display-name (required)
 :: The display name of the agent, intended to be displayed to a user by the
-     requester. The requester should indicate through the UI if the device
+     requester. The requester should indicate through the UI if the responder
      is not authenticated or if the display name changes.
 
 : model-name (optional)
@@ -320,13 +320,13 @@ The agent-info-response messages contain the following properties:
     the device.  This is used mainly for debugging purposes, but may be
     displayed to the user of the requesting agent.
 
-: supports-audio (optional)
-:: The agent has to indicate that it supports audio streaming. If false
-    or not included, it is assumed audio streaming is not supported.
+: receives-audio (optional)
+:: The agent has to indicate that it supports audio. If false
+    or not included, it is assumed audio content is not supported.
 
-: supports-video (optional)
-:: The agent has to indicate that it supports video streaming. If false
-    or not included, it is assumed video streaming is not supported.
+: receives-video (optional)
+:: The agent has to indicate that it supports video . If false
+    or not included, it is assumed video content is not supported.
 
 
 Listening agents act as QUIC clients.  Advertising agents act as QUIC servers.
@@ -349,7 +349,7 @@ If a client agent wishes to send messages to a server agent, the client
 agent can connect to the server agent "on demand"; it does not need to
 keep the connection alive.
 
-The agent-info-response message and agent-status-response
+The agent-info-response and agent-status-response
 messages may be extended to include additional information not defined
 in this spec.  If done ad-hoc by applications and not in future specs,
 keys should be chosen to avoid collision, such as by choosing large
--- a/index.bs
+++ b/index.bs
@@ -151,7 +151,7 @@ Presentation API Requirements {#requirements-presentation-api}
     {{PresentationConnection}} to an existing presentation on the
     receiver, given its [=presentation request URL=] and [=presentation ID=].
 
-6.  It must be possible to to close a {{PresentationConnection}} between a
+6.  It must be possible to close a {{PresentationConnection}} between a
     controller and a presentation, and signal both parties with the
     reason why the connection was closed.
 
@@ -297,32 +297,85 @@ mechanism for metadata discovery.
 Transport and metadata discovery with QUIC {#transport}
 =======================================================
 
-If a listening agent wants to connect to an advertising agent, or to
-learn further metadata about it, it initiates a [[QUIC]] connection to
-the IP and port from the SRV record.  Prior to authentication, a message may be
-exchanged (such as further metadata), but such info should be treated as
-unverified (such as indicating to a user that a display name of an
+If a listening agent ("listener") wants to connect to an advertising agent
+("advertiser"), or to learn further metadata about it, it initiates a [[QUIC]]
+connection to the IP and port from the SRV record.  Prior to authentication,
+a message may be exchanged (such as further metadata), but such info should be
+treated as unverified (such as indicating to a user that a display name of an
 unauthenticated agent is unverified).
 
-To learn further metadata, an agent may send an agent-info-request
-message (see [[#appendix-a]]) and receive back an agent-info-response message.  The
-messages may contain the following information with the following meaning:
+To learn further metadata, the listener may send an agent-info-request
+message (see [[#appendix-a]]) and receive back an agent-info-response message.
+The messages may contain the following properties:
 
 : display-name (required)
-:: The display name of the responding agent intended
-    to be displayed to a user by the requesting agent.  If the responding agent
-    is not yet authenticated, the requesting agent should make UI affordance for
-    indicating to the user that the display name is not yet verified.  If the
-    responding agent changes its display name, the requesting agent should
-    make UI affordance for indicating to the user that the display name has
-    changed.
+:: The display name of the advertiser, intended to be displayed to a user by the
+     listener. The listener should indicate through the UI if the advertiser
+     is not authenticated or if the display name changes.
 
 : model-name (optional)
-:: If the agent is a hardware device, the model name of
+:: If the advertiser is a hardware device, the model name of
     the device.  This is used mainly for debugging purposes, but may be
     displayed to the user of the requesting agent.
 
-<!-- TODO: Add device type and/or capabilities -->
+: audio-only (optional)
+:: Generally speaking, advertisers are assumed to support streaming both
+    audio and video. If a advertiser wishes to indicate that they only support audio
+    stream, they can pass this property as true. In that case, the listener
+    should indicate through the UI that the advertiser is audio only.
+
+If the advertiser is already authenticated, the listener has the ability to
+request additional information by sending an streaming-capabilities-request
+message, and receive back a streaming-capabilities-response message with the
+following properties:
+
+: max-pixel-fill-rate (required)
+:: The advertiser shall reply with the supported pixel fill rate in megapixels per
+    second. Requesting agents can use this information to make better-educated
+    decisions about what resolution and frame rates to stream content at.
+
+: supported-codecs (required)
+:: The advertiser must indicate to the listener what codecs it understands for
+    remote playback.
+
+: min-resolution (optional)
+:: The advertiser may indicate what it considers its minimum resolution to be, for
+    setting a minimum boundary on content quality.
+
+: max-resolution (optional)
+:: The advertiser may indicate the maximum resolution and frame rate. Note that
+    this may not be the ideal, depending on the max-pixel-fill-rate.
+
+: max-audio-channels (optional)
+:: The advertiser may indicate the maximum number of audio channels in abilities
+    standard audio setup, e.g. 2 for stereo, 5 or 7 for a surround setup.
+
+: audio-bit-rate-limits (optional)
+:: The advertiser may indicate the practical audio bit rate limits, including the
+    minimum and maximum bit rates, and the maximum latency. Physical hardware
+    requirements, such as maximum decoder throughput or maximum audio buffer
+    size inform these values.
+
+: video-bit-rate-limits (optional)
+:: The advertiser may indicate the practical audio bit rate limits, including the
+    minimum and maximum bit rates, and the maximum latency. Physical hardware
+    requirements, such as maximum decoder throughput or maximum video buffer
+    size inform these values.
+
+: aspect-ratio (optional)
+:: The advertiser may indicate what its ideal aspect ratio is, e.g. a 16:10
+    display could return this value as 1.6 to indicate its preferred content
+    scaling.
+
+: color-profiles (optional)
+:: The advertiser may indicate what color profiles are understood, such as
+    sRGB v4. The listener may use these values to determine how to encode
+    video.
+
+: supports-scaling (optional)
+:: If omitted, this value is assumed to be false. Setting to "True" indicates
+    that the advertiser can scale content that is of an unsupported resolution or
+    non-ideal aspect ratio. If false, the listener must do the scaling.
 
 Listening agents act as QUIC clients.  Advertising agents act as QUIC servers.
 
@@ -339,7 +392,6 @@ frames should not be used.  An idle_timeout transport parameter of 25 seconds is
 recommended.  The agent should behave as though a timer less than the
 idle_timeout were reset every time a message is sent on a QUIC stream.  If the
 timer expires, a agent-status-request message should be sent.
-
 
 If a client agent wishes to send messages to a server agent, the client
 agent can connect to the server agent "on demand"; it does not need to
--- a/index.bs
+++ b/index.bs
@@ -302,41 +302,44 @@ If a listening agent ("listener") wants to connect to an advertising agent
 connection to the IP and port from the SRV record.  Prior to authentication,
 a message may be exchanged (such as further metadata), but such info should be
 treated as unverified (such as indicating to a user that a display name of an
-unauthenticated agent is unverified).
+unauthenticated agent is unverified). Note that both listeners and advertisers
+may request information about each other (performing as a "requester" to
+query a "requestee").
 
-To learn further metadata, the listener may send an agent-info-request
+To learn further metadata, devices may send an agent-info-request
 message (see [[#appendix-a]]) and receive back an agent-info-response message.
+Note that both the listener and advertiser may send this request to learn about
+the capabilities of the other device.
+
 The messages may contain the following properties:
 
 : display-name (required)
-:: The display name of the advertiser, intended to be displayed to a user by the
-     listener. The listener should indicate through the UI if the advertiser
+:: The display name of the requestee, intended to be displayed to a user by the
+     requester. The requester should indicate through the UI if the device
      is not authenticated or if the display name changes.
 
 : model-name (optional)
-:: If the advertiser is a hardware device, the model name of
+:: If the requestee is a hardware device, the model name of
     the device.  This is used mainly for debugging purposes, but may be
     displayed to the user of the requesting agent.
 
-: audio-only (optional)
-:: Generally speaking, advertisers are assumed to support streaming both
-    audio and video. If a advertiser wishes to indicate that they only support audio
-    stream, they can pass this property as true. In that case, the listener
-    should indicate through the UI that the advertiser is audio only.
+: supports-audio (optional)
+:: The requestee has to indicate that it supports audio streaming. If false
+    or not included, it is assumed audio streaming is not supported.
+
+: supports-video (optional)
+:: The requestee has to indicate that it supports video streaming. If false
+    or not included, it is assumed video streaming is not supported.
 
 If the advertiser is already authenticated, the listener has the ability to
 request additional information by sending an streaming-capabilities-request
 message, and receive back a streaming-capabilities-response message with the
 following properties:
 
-: max-pixel-fill-rate (required)
-:: The advertiser shall reply with the supported pixel fill rate in megapixels per
-    second. Requesting agents can use this information to make better-educated
-    decisions about what resolution and frame rates to stream content at.
-
-: supported-codecs (required)
+: codecs (required)
 :: The advertiser must indicate to the listener what codecs it understands for
-    remote playback.
+    remote playback. Note that the codec includes both (1) the codec name, and
+    (2) the max pixels per second supported for that codec.
 
 : min-resolution (optional)
 :: The advertiser may indicate what it considers its minimum resolution to be, for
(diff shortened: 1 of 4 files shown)


index.bs
```diff
@@ -843,4 +1380,227 @@ result = (
   unknown-error: 199
 )
 
+; type key 17
+remote-playback-availability-request = {
+  request
+  1: [* text] ; urls
+  2: microseconds ; watch-duration
+  3: int ; watch-id
```
rwkeane: Seems be better to use `uint`instead of `int` for consistency with `connection-id`, `request-id`

index.bs
```diff
@@ -843,4 +1380,227 @@ result = (
...
+}
+
+; type key 18
+remote-playback-availability-response = {
+  response
+  url-availabilities: [* url-availability]
```
rwkeane: integer key to minimize payload

index.bs
```diff
@@ -843,4 +1380,227 @@ result = (
...
+    loading: 2
+    no-source: 3
+  ) ; loading
+  ? 4: &(
+    nothing: 0
+    metdata: 1
+    current: 2
+    future: 3
+    enough: 4
+  ) ; loaded
```
rwkeane: Even if `3` and `4`'s types aren't used elsewhere, I think pulling them out into enums defined outside this map would be more readable. Plus the generated C++ code would be more readable and easier to use

index.bs
```diff
@@ -843,4 +1380,227 @@ result = (
...
+}
+
+added-text-track = {
+  1: &(
+    subtitles
+    captions
+    descriptions
+    chapters
+    metadata
+  ) ; kind
```
rwkeane: same as above - define the enum as a type outside the map

index.bs
```diff
@@ -843,4 +1380,227 @@ result = (
...
+    decode-error: 3
+    source-not-supported: 4
+  )
+  message: text
+]
+
+track-state = (
+  1: text ; id
+  2: text ; label
+  3: text ; language
```
rwkeane: Is language shared with `video-track-state` too?

index.bs
```diff
@@ -843,4 +1380,227 @@ result = (
...
+)
+
+audio-track-state = {
+  track-state
+  4: bool ; enabled
+}
+
+video-track-state = {
+  track-state
+  5 : bool ; selected
```
rwkeane: looks like this should be `4` not `5`

index.bs
```diff
@@ -843,4 +1380,227 @@ result = (
...
+}
+
+track-track-state = {
+  track-state
+  6 : mode ; text-track-mode
```
rwkeane: looks like this should be `4` not `6`

index.bs
```diff
@@ -843,4 +1380,227 @@ result = (
...
+}
+
+; type key 118
+remote-playback-termination-response = {
+  response
+  1: &result ; result
+}
+
+; type key 119
+remote-playback-termination-event = {
```
rwkeane: Do we still need ` 0: request-id ; request-id` added here, even if isn't a request or response?

index.bs
```diff
@@ -843,4 +1380,227 @@ result = (
...
+  ? 22: [* text-track-state] text-tracks
+}
+
+added-text-track = {
+  1: &(
+    subtitles
+    captions
+    descriptions
+    chapters
+    metadata
```
rwkeane: Give integer values

index.bs
btolsch: This comment and your others apply to Peter's PR (#134).  Unfortunately, it seems like GitHub doesn't have support for dependent PRs (or maybe just not from separate authors?).  Looking at the individual commits is easier for me than looking at the total diff.

index.bs
```diff
@@ -284,31 +284,55 @@ mechanism for metadata discovery.
 Transport and metadata discovery with QUIC {#transport}
 =======================================================
 
-If a listening agent wants to connect to an advertising agent, or to
-learn further metadata about it, it initiates a [[QUIC]] connection to
-the IP and port from the SRV record.  Prior to authentication, a message may be
-exchanged (such as further metadata), but such info should be treated as
-unverified (such as indicating to a user that a display name of an
+If a listening agent ("controller") wants to connect to an advertising agent
```
btolsch: I'm sure Peter and/or Mark will chime in, but I think they have avoided controller/receiver terminology is because it seems Presentation-API-centric.  I'm neutral about this though.

index.bs
```diff
@@ -284,31 +284,55 @@ mechanism for metadata discovery.
...
 
 : model-name (optional)
-:: If the agent is a hardware device, the model name of
+:: If the receiver is a hardware device, the model name of
     the device.  This is used mainly for debugging purposes, but may be
     displayed to the user of the requesting agent.
 
+: audio-only (optional)
+:: Generally speaking, receivers are assumed to support streaming both
+audio and video. If an receiver wishes to indicate that they only support audio
```
btolsch: s/an receiver/a receiver/

aside: Oh man, did GitHub remove the suggestion feature?  That was really nice for typos like this.

index.bs
```diff
@@ -284,31 +284,55 @@ mechanism for metadata discovery.
...
+following properties:
+
+: supported-resolutions (required)
+:: The receiver shall indicate what resolutions and refresh rates it supports
+streaming content at.
+
+<!-- TODO(jophba): figure out the units for this -->
+: max-pixel-fill-rate (optional)
+:: In combination with the supported resolutions, the receiver has the
+option to to reply with the supported pixel fill rate in <UNITS>.
```
btolsch: s/to to/to/

index.bs
```diff
@@ -1206,11 +1198,18 @@ resolution = {
   2: [1* float64] ; refresh-rates
 }
 
-<!-- TODO(jophba): what codecs do we support? -->
-<!-- TODO(jophba): do we need this? -->
-<!-- TODO(jophba): flag enum? -->
 codec = (
-
+  H-264-1080
```
btolsch: These need to be `name: number`.

index.bs
```diff
@@ -1206,11 +1198,18 @@ resolution = {
...
+  H-264-720
```
btolsch: Why is there a separate 720 and 1080 h264?

index.bs
baylesj: Thx

index.bs
baylesj: I'll leave as is for now, and bring it up with them.

index.bs
```diff
@@ -141,7 +141,7 @@ Presentation API Requirements {#requirements-presentation-api}
     {{PresentationConnection}} to an existing presentation on the
     receiver, given its [=presentation request URL=] and [=presentation ID=].
 
-6.  It must be possible to to close a {{PresentationConnection}} between a
```
pthatcherg: About all these typos you found and fixed: could you make a separate PR that fixes those?  It would be nice to have them fixed independently of this PR.

index.bs
pthatcherg: I like controller/receiver as terms in general, but in the context of mDNS and discovery, we can't assume the receiver is always the advertiser and the controller is always the listener.

index.bs
```diff
@@ -287,32 +287,85 @@ mechanism for metadata discovery.
...
+("receiver"), or to learn further metadata about it, it initiates a [[QUIC]]
+connection to the IP and port from the SRV record.  Prior to authentication,
+a message may be exchanged (such as further metadata), but such info should be
+treated as unverified (such as indicating to a user that a display name of an
 unauthenticated agent is unverified).
 
-To learn further metadata, an agent may send an agent-info-request
-message (see [[#appendix-a]]) and receive back an agent-info-response message.  The
-messages may contain the following information with the following meaning:
+To learn further metadata, the controller may send an agent-info-request
```
pthatcherg: Actually, both sides will probably send this message to learn about the other side.

index.bs
```diff
@@ -287,32 +287,85 @@ mechanism for metadata discovery.
...
-    responding agent changes its display name, the requesting agent should
-    make UI affordance for indicating to the user that the display name has
-    changed.
+:: The display name of the receiver, intended to be displayed to a user by the
+     controller. The controller should indicate through the UI if the receiver
+     is not authenticated or if the display name changes.
 
 : model-name (optional)
-:: If the agent is a hardware device, the model name of
+:: If the receiver is a hardware device, the model name of
```
pthatcherg: The device might be a controller.  Imagine a TV wants to connect back to controller devices.

index.html
```diff
@@ -1696,30 +1698,85 @@ <h2 class="heading settled" data-level="3" id="discovery"><span class="secno">3.
...
+a message may be exchanged (such as further metadata), but such info should be
+treated as unverified (such as indicating to a user that a display name of an
 unauthenticated agent is unverified).</p>
-   <p>To learn further metadata, an agent may send an agent-info-request
-message (see <a href="#appendix-a">Appendix A: Messages</a>) and receive back an agent-info-response message.  The
-messages may contain the following information with the following meaning:</p>
+   <p>To learn further metadata, the controller may send an agent-info-request
+message (see <a href="#appendix-a">Appendix A: Messages</a>) and receive back an agent-info-response message.
+The messages may contain the following properties:</p>
    <dl>
```
pthatcherg: Same here: we can't assume the agent is a controller.

index.html
```diff
@@ -1540,7 +1540,9 @@ <h3 class="heading settled" data-level="1.1" id="terminology"><span class="secno
...
+a browser). <a data-link-type="dfn" href="https://w3c.github.io/presentation-api/#dfn-presentation-display-availability" id="ref-for-dfn-presentation-display-availability">presentation display availability</a> refers to whether or not a <a data-link-type="dfn" href="https://w3c.github.io/presentation-api/#dfn-receiver" id="ref-for-dfn-receiver①">receiver</a> is compatible with a <a data-link-type="dfn" href="https://w3c.github.io/presentation-api/#dfn-presentation-request-url" id="ref-for-dfn-presentation-request-url">presentation request URL</a>.  However, in the <a data-link-type="biblio" href="#biblio-presentation-api">Presentation API</a>, a "controller" refers to as a specific
+browsing context within the browser, whereas here the "controller" refers to the
+browser itself, although it may be acting on behalf of a browsing context.</p>
    <p>We borrow terminology from the <a data-link-type="biblio" href="#biblio-remote-playback">Remote Playback API</a>.  The
```
pthatcherg: Hmm... do you need to rebase again?  I'm seeing in your diff things I made in the last patch of my PR.

index.html
```diff
@@ -1696,30 +1698,85 @@ <h2 class="heading settled" data-level="3" id="discovery"><span class="secno">3.
...
-responding agent changes its display name, the requesting agent should
-make UI affordance for indicating to the user that the display name has
-changed.</p>
+     <p>The display name of the receiver, intended to be displayed to a user by the
+ controller. The controller should indicate through the UI if the receiver
+ is not authenticated or if the display name changes.</p>
     <dt data-md>model-name (optional)
     <dd data-md>
-     <p>If the agent is a hardware device, the model name of
+     <p>If the receiver is a hardware device, the model name of
```
pthatcherg: Here again with assuming receiver == hardware

index.html
```diff
@@ -1696,30 +1698,85 @@ <h2 class="heading settled" data-level="3" id="discovery"><span class="secno">3.
...
     <dd data-md>
-     <p>The display name of the responding agent intended
-to be displayed to a user by the requesting agent.  If the responding agent
-is not yet authenticated, the requesting agent should make UI affordance for
-indicating to the user that the display name is not yet verified.  If the
-responding agent changes its display name, the requesting agent should
-make UI affordance for indicating to the user that the display name has
-changed.</p>
+     <p>The display name of the receiver, intended to be displayed to a user by the
+ controller. The controller should indicate through the UI if the receiver
```
pthatcherg: Either side can get the display name of other; it's mutual.  It's not just receiver -> controller.

index.bs
```diff
@@ -1527,10 +1579,64 @@ agent-info-response = {
...
+  eAC3: 6
+  FLAC: 7
+  PCM: 8
+  Vorbis: 9
+)
+
+bit-rate-limits = {
+ 0: uint ; min-bit-rate
+ 1: uint ; max-bit-rate
+ 2: uint ; max-latency
```
pthatcherg: I don't think these should be bundled together.  It's much more common to want a max than a min bitrate.

pthatcherg: max-latency?  I don't know what that means and it seems very hard to define.  Perhaps we can address something like the sender being able to control the receiver's jitter buffer, but it doesn't seem like a capability at all.

index.bs
baylesj: %s/controller/listener/g
%s/receiver/advertiser/g

index.html
baylesj: I think the PR's diff is just bunk, the log looks right to me.

index.bs
baylesj: I'll cleanup language.

index.bs
baylesj: Deleting.

index.html
```diff
@@ -2351,10 +2790,64 @@ <h2 class="heading settled" id="appendix-a"><span class="content">Appendix A: Me
...
+  FLAC: 7
+  PCM: 8
+  Vorbis: 9
+)
+
+bit-rate-limits = {
+ 0: uint ; min-bit-rate
+ 1: uint ; max-bit-rate
+ 2: uint ; max-latency
+}
```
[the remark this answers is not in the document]
baylesj: I'll split up.

index.bs
```diff
@@ -287,32 +287,85 @@ mechanism for metadata discovery.
...
+
+: max-audio-channels (optional)
+:: The receiver may indicate the maximum number of audio channels in abilities
+    standard audio setup, e.g. 2 for stereo, 5 or 7 for a surround setup.
+
+: audio-bit-rate-limits (optional)
+:: The receiver may indicate the practical audio bit rate limits, including the
+    minimum and maximum bit rates, and the maximum latency. Physical hardware
+    requirements, such as maximum decoder throughput or maximum audio buffer
+    size inform these values.
```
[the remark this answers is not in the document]
baylesj: Removing audio bit rate.

index.bs
```diff
@@ -1527,10 +1579,64 @@ agent-info-response = {
...
+
+bit-rate-limits = {
+ 0: uint ; min-bit-rate
+ 1: uint ; max-bit-rate
+ 2: uint ; max-latency
+}
+
+streaming-capabilities = {
+  0: uint ; max-pixel-fill-rate
+  1: [1* codec] ; supported-codecs
```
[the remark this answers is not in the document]
baylesj: Changing name.

index.html
```diff
@@ -2351,10 +2790,64 @@ <h2 class="heading settled" id="appendix-a"><span class="content">Appendix A: Me
...
+
+streaming-capabilities = {
+  0: uint ; max-pixel-fill-rate
+  1: [1* codec] ; supported-codecs
```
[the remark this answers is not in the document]
baylesj: String sounds great.

index.bs
```diff
@@ -287,32 +287,85 @@ mechanism for metadata discovery.
...
+
+If the receiver is already authenticated, the controller has the ability to
+request additional information by sending an streaming-capabilities-request
+message, and receive back a streaming-capabilities-response message with the
+following properties:
+
+: max-pixel-fill-rate (required)
+:: The receiver shall reply with the supported pixel fill rate in megapixels per
+    second. Requesting agents can use this information to make better-educated
+    decisions about what resolution and frame rates to stream content at.
```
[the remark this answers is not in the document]
baylesj: Added new codec type.

index.bs
```diff
@@ -287,32 +287,85 @@ mechanism for metadata discovery.
...
+: audio-only (optional)
+:: Generally speaking, receivers are assumed to support streaming both
+    audio and video. If a receiver wishes to indicate that they only support audio
+    stream, they can pass this property as true. In that case, the controller
+    should indicate through the UI that the receiver is audio only.
+
+If the receiver is already authenticated, the controller has the ability to
+request additional information by sending an streaming-capabilities-request
+message, and receive back a streaming-capabilities-response message with the
+following properties:
```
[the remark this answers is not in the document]
baylesj: I'll push a new revision addressing your feedback, then we can pull over.

index.bs
```diff
@@ -287,32 +287,85 @@ mechanism for metadata discovery.
...
+:: If the receiver is a hardware device, the model name of
     the device.  This is used mainly for debugging purposes, but may be
     displayed to the user of the requesting agent.
 
-<!-- TODO: Add device type and/or capabilities -->
+: audio-only (optional)
+:: Generally speaking, receivers are assumed to support streaming both
+    audio and video. If a receiver wishes to indicate that they only support audio
+    stream, they can pass this property as true. In that case, the controller
+    should indicate through the UI that the receiver is audio only.
```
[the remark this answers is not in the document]
baylesj: Agreed.

index.bs
```diff
@@ -287,32 +287,85 @@ mechanism for metadata discovery.
...
+:: The receiver may indicate the practical audio bit rate limits, including the
+    minimum and maximum bit rates, and the maximum latency. Physical hardware
+    requirements, such as maximum decoder throughput or maximum audio buffer
+    size inform these values.
+
+: video-bit-rate-limits (optional)
+:: The receiver may indicate the practical audio bit rate limits, including the
+    minimum and maximum bit rates, and the maximum latency. Physical hardware
+    requirements, such as maximum decoder throughput or maximum video buffer
+    size inform these values.
```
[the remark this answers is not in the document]
baylesj: Removed audio bit rate limits.

index.bs
baylesj: https://github.com/webscreens/openscreenprotocol/pull/151

index.bs
```diff
@@ -297,32 +297,84 @@ mechanism for metadata discovery.
...
-unverified (such as indicating to a user that a display name of an
-unauthenticated agent is unverified).
+If a listening agent ("listener") wants to connect to an advertising agent
+("advertiser"), or to learn further metadata about it, it initiates a [[QUIC]]
+connection to the IP and port from the SRV record.  Prior to authentication,
+a message may be exchanged (such as further metadata), but such info should be
+treated as unverified (such as indicating to a user that a display name of an
+unauthenticated agent is unverified). Note that both listeners and advertisers
+may request information about each other (performing as a "requester" to
+query a "requestee").
```
mfoltzgoogle: This paragraph is vague and confusing.  It starts by suggesting that the listener wants to learn about the advertiser.  But then says that each side can learn about the other once they are connected.   It's also unclear what message in the spec can be sent as "unverified" data.

I would suggest dropping the last sentence as it's covered in a more specific way in the next paragraph.

index.bs
```diff
@@ -297,32 +297,84 @@ mechanism for metadata discovery.
...
 
-To learn further metadata, an agent may send an agent-info-request
-message (see [[#appendix-a]]) and receive back an agent-info-response message.  The
-messages may contain the following information with the following meaning:
+To learn further metadata, devices may send an agent-info-request
+message (see [[#appendix-a]]) and receive back an agent-info-response message.
+Note that both the listener and advertiser may send this request to learn about
```
mfoltzgoogle: s/listener and advertiser/any agent/

index.bs
```diff
@@ -297,32 +297,84 @@ mechanism for metadata discovery.
...
+the capabilities of the other device.
+
+The messages may contain the following properties:
```
mfoltzgoogle: s/may// since there are required properties

mfoltzgoogle: Are these properties of the agent-info-request or the agent-info-response?

index.bs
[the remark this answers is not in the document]
mfoltzgoogle: codecs could be used for streaming or remote playback, so I think we should keep it generic.  Unless we believe that the capabilities will be different for a given device.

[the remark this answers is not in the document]
baylesj: Removed streaming capabilities, will fixup other patch.

index.bs
```diff
@@ -287,32 +287,85 @@ mechanism for metadata discovery.
...
+: max-pixel-fill-rate (required)
+:: The receiver shall reply with the supported pixel fill rate in megapixels per
+    second. Requesting agents can use this information to make better-educated
+    decisions about what resolution and frame rates to stream content at.
+
+: supported-codecs (required)
+:: The receiver must indicate to the controller what codecs it understands for
+    remote playback.
+
+: min-resolution (optional)
```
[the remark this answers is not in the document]
mfoltzgoogle: +1  I'm not sure codecs belong in a device-level metadata, but instead should be part the negotiation specifically for streaming or remote playback.  Ideally we can share some messages between the two because there's overlap in the use case: the controller has some media it wants to play on the receiver with a given codec, resolution/framerate, and bitrate.  The receiver can answer if it can play that media at all and/or play it smoothly or efficiently (see the [Media Capabilities API](https://wicg.github.io/media-capabilities/) for a Web API for querying this data).

index.bs
```diff
@@ -287,32 +287,85 @@ mechanism for metadata discovery.
...
+: video-bit-rate-limits (optional)
+:: The receiver may indicate the practical audio bit rate limits, including the
+    minimum and maximum bit rates, and the maximum latency. Physical hardware
+    requirements, such as maximum decoder throughput or maximum video buffer
+    size inform these values.
+
+: aspect-ratio (optional)
+:: The receiver may indicate what its ideal aspect ratio is, e.g. a 16:10
+    display could return this value as 1.6 to indicate its preferred content
+    scaling.
```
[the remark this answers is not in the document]
mfoltzgoogle: Nit, but this feels better as a rational number, because of 16:9, 4:3, etc.

index.bs
```diff
@@ -287,32 +287,85 @@ mechanism for metadata discovery.
...
+
+: color-profiles (optional)
+:: The receiver may indicate what color profiles are understood, such as
+    sRGB v4. The controller may use these values to determine how to encode
+    video.
+
+: supports-scaling (optional)
```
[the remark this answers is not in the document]
mfoltzgoogle: In general there are multiple native resolutions that have hardware support for scaling.  The receiver could list a set of supported resolutions and then indicate if non-standard resolutions are also supported.

index.bs
```diff
@@ -297,32 +297,40 @@ mechanism for metadata discovery.
...
 
-<!-- TODO: Add device type and/or capabilities -->
+: supports-audio (optional)
+:: The requestee has to indicate that it supports audio streaming. If false
+    or not included, it is assumed audio streaming is not supported.
+
+: supports-video (optional)
+:: The requestee has to indicate that it supports video streaming. If false
+    or not included, it is assumed video streaming is not supported.
+
```
mfoltzgoogle: In the future we should consider capabilities for audio input and video input as well.  But these capabilities are more often associated with removable peripherals, so it may not be suitable for device-level metadata.

Maybe we should make all capabilities playback-level metadata?  That is, part of the url-availability-request, media-request and/or remote-playback-availability-request?

index.bs
baylesj: Clarified.

index.bs
baylesj: Agreed.

index.bs
baylesj: Changed.

index.bs
baylesj: That's a good question. I'll leave this for now, I'll be in Fremont tomorrow/Friday if you want to sync up in person and hammer down some of the specifics around this.

rektide: This is an even bigger can of worms, but I'd like to see input device support as a capability too. The trendy new big touchscreen collaborative displays come to mind as the sort of thing one would expect to openscreenprotocol into, which would have their own input device systems.

I don't know whether this capability ought be broken down into smaller pieces- `supports-input` or `supports-touch` `supports-keyboard` `supports-mouse`?

As per previous comments, maybe doesn't need to be done now? I can open a follow-up item depending on peoples feels here.

baylesj: I would vote for a follow up item, personally. I generally prefer smaller patches when possible.

index.bs
```diff
@@ -297,32 +297,37 @@ mechanism for metadata discovery.
...
+a display name of an unauthenticated agent is unverified).
 
 To learn further metadata, an agent may send an agent-info-request
-message (see [[#appendix-a]]) and receive back an agent-info-response message.  The
-messages may contain the following information with the following meaning:
+message (see [[#appendix-a]]) and receive back an agent-info-response message.
+Any agent may send this request to learn about the capabilities of
+another device.
+
+The agent-info-response messages contain the following properties:
```
pthatcherg: "The X message contains" or "X messages contain", not "The X messages contain,"

index.bs
```diff
@@ -297,32 +297,37 @@ mechanism for metadata discovery.
...
 : display-name (required)
-:: The display name of the responding agent intended
-    to be displayed to a user by the requesting agent.  If the responding agent
-    is not yet authenticated, the requesting agent should make UI affordance for
-    indicating to the user that the display name is not yet verified.  If the
-    responding agent changes its display name, the requesting agent should
-    make UI affordance for indicating to the user that the display name has
-    changed.
+:: The display name of the agent, intended to be displayed to a user by the
+     requester. The requester should indicate through the UI if the device
```
pthatcherg: "if the device" => "if the responder"

index.bs
```diff
@@ -297,32 +297,37 @@ mechanism for metadata discovery.
...
+     is not authenticated or if the display name changes.
 
 : model-name (optional)
 :: If the agent is a hardware device, the model name of
     the device.  This is used mainly for debugging purposes, but may be
     displayed to the user of the requesting agent.
 
-<!-- TODO: Add device type and/or capabilities -->
+: supports-audio (optional)
+:: The agent has to indicate that it supports audio streaming. If false
```
pthatcherg: It think we should show support for audio and video in general, not specific to streaming (such as for the presentation API).  This is mostly to know if we should show an a/v icon or just an audio icon.  

Also, the name should probably be "receives-audio" and "receives-video".

baylesj: Agreed.

index.bs
baylesj: +1

index.bs
baylesj: +1

baylesj:
@mfoltzgoogle, thoughts on closed comments below. TL;DR I believe I have addressed all of your comments in my other patch,
https://github.com/webscreens/openscreenprotocol/pull/154

**https://github.com/webscreens/openscreenprotocol/pull/140#discussion_r272304956**
"Codecs should be generic":
_Agreed, cleaned up in second patch._

**https://github.com/webscreens/openscreenprotocol/pull/140#discussion_r272308357**
"Codecs belong in streaming":
_Agreed, moved as part of moving to second patch._

**https://github.com/webscreens/openscreenprotocol/pull/140#discussion_r272308928**
"Rational number":
_Makes sense. Added a new ratio type, since we don't appear to have one._

**https://github.com/webscreens/openscreenprotocol/pull/140#discussion_r272309518**
"Multiple native resolutions":
_Agreed, I updated to add a : native-resolutions (optional) field._

mfoltzgoogle:
Approved the current PR with two items for follow-up discussion:

- In the `agent-info` message, using booleans versus a set-of-enums to describe capabilities
- How to describe the semantics of `receives-audio` and `receives-video` more specifically.  Does that mean an agent is required to process set of messages, a set of codecs or is it entirely up to the implementation?

Still awaiting approval from @pthatcherg.

pthatcherg:
I think enums instead of bools is an interesting idea.  

As for "receives-audio" and "receives-video", I think the part we're missing is to have a capabilities for indicating what parts of the protocol the agent speaks.  For example, it acts as a receiver for presentation API or remote playback API or streaming or all of them.  Or it acts as a sender/controller for presentation API or remote playback API or streaming or all of them.

I think "receives-audio" and "receives-video" implies that it's a receiver, and for now that probably means both presentation API and remote playback API, but not streaming.

pthatcherg:
I think we can land this and continue discussing in more detail.

PR #154, opened 2019-04-04
Flesh out streaming capabilities

baylesj:


baylesj:
May have to do some fancy things around cleaning up the diff. I updated so that the CDDL stuff is merged appropriately.

baylesj:
@pthatcherg, this patch currently contains only the stuff from my capabilities PR, with edits and improvements. I'm thinking we can review this one, jam it into your patch, and do the other sections separately?

index.bs
```diff
@@ -1224,52 +1224,80 @@ as the *sender* and the agent receiving the media as the *receiver*.
...
 --------------------------------------------
+If the advertiser is already authenticated, the requester has the ability to
+request additional information by sending an streaming-capabilities-request
+message, and receive back a streaming-capabilities-response message with the
+following properties:
 
-<!--
-  TODO:
+: receive-audio (required)
+:: A list of audio capabilities. For an explanation of fields, see below.
```
pthatcherg: Perhaps "a list of capabilities for receiving audio".  Here and similar below.  Some day, we may have capabilites for sending as well, so I want to make it clear everywhere that these are receiving capabilities.

index.bs
```diff
@@ -1224,52 +1224,80 @@ as the *sender* and the agent receiving the media as the *receiver*.
...
-  Put in something like this:
+: receive-video (required)
+:: A list of video capabilities. For an explanation of fields, see below.
 
-streaming-capabilities-request {
-  request
-}
+: receive-data (required)
+:: A list of arbitrary data formats (represented as codecs) the device supports
+    for receiving data.
```
pthatcherg: I think I'd rather call codecs formats than call data formats codecs.  It's actually more precise as well, since codecs can be in different profiles/modes.  It's just that we commonly say "codec" instead of "format".

I'll give more explicit suggestions down below.

index.bs
```diff
@@ -1224,52 +1224,80 @@ as the *sender* and the agent receiving the media as the *receiver*.
...
-}
+The codec type is used as the basis for audio, video, and data capabilities.
+Codes are composed of the following properties:
 
-streaming-capabilities = {
-  receive-audio: [* receive-audio-capability]
-  receive-video: [* receive-video-capability]
-  receive-data: [* receive-data-capability]
-}
+: codec-name (required)
```
pthatcherg: Perhaps just call this "name" and give examples of expected values: "vp8", "h264", "opus"

index.bs
```diff
@@ -1224,52 +1224,80 @@ as the *sender* and the agent receiving the media as the *receiver*.
...
+:: The name of the codec.
 
+: parameters (required)
+:: A list of (key, value) parameters that can be used to transmit arbitrary
+    information about a codec.
```
pthatcherg: Should say something about these being for codec-specific parameters that aren't applicable to all codecs of a certain media type (audio, video, data).

index.bs
```diff
@@ -2375,12 +2402,64 @@ data-frame = {
...
+  1: streaming-capabilities ; streaming-capabilities
+}
+
+streaming-capabilities = {
+  0: [* receive-audio-capability] ; receive-audio
+  1: [* receive-video-capability] ; receive-video
+  2: [* receive-data-capability] ; receive-data
+}
+
+codec = {
```
pthatcherg: We could just make this a "format" as the type and have the field name for audio and video be "codec" so ".codec.name" and ".codec.parameters" works

index.bs
```diff
@@ -1224,52 +1224,80 @@ as the *sender* and the agent receiving the media as the *receiver*.
...
-  codec-parameters: [* codec-parameter]
-  ? max-resolution: video-resolution
-  ? max-frames-per-second: float
-  ? max-pixels-per-second: float
-}
+: max-audio-channels (optional)
+:: An optional field indicating the maximum amount of audio
+    channels the device is capable of supporting. For example, a stereo speaker
+    system could supply this optional field with a value of "2" for the left and
+    right channels.
```
pthatcherg: What's the default?  1 or 2?  I'd to with 2.

index.bs
```diff
@@ -1224,52 +1224,80 @@ as the *sender* and the agent receiving the media as the *receiver*.
...
 
-receive-data-capability = {
-  format: text ; we define "text-cue" for text cues
-  format-parameters: codec-parameters
-}
+: min-bit-rate (optional)
+:: An optional field indicating the minimum audio bit rate that
+    the device can handle.
```
pthatcherg: Should point out default is no minimum.

index.bs
```diff
@@ -1224,52 +1224,80 @@ as the *sender* and the agent receiving the media as the *receiver*.
...
-}
+Video capabilities are similarly composed of the above codec type, with the
+following additional fields:
 
--->
+: max-resolution (optional)
+:: An optional field composed of a resolution, which includes (1) an unsigned
+    integer indicating horizontal resolution, and (2) an unsigned integer
+    indicating vertical resolution. The device can pass this field to indicate
+    the maximum resolution it supports.
```
pthatcherg: Default is no max?

index.bs
```diff
@@ -1224,52 +1224,80 @@ as the *sender* and the agent receiving the media as the *receiver*.
...
+
+: max-frames-per-second (optional)
+:: An optional field indicating the maximum frames-per-second the device is
+    capable of processing.
```
pthatcherg: Default is no max?

index.bs
```diff
@@ -1224,52 +1224,80 @@ as the *sender* and the agent receiving the media as the *receiver*.
...
+:: An optional field indicating the maximum amount of audio
+    channels the device is capable of supporting. For example, a stereo speaker
+    system could supply this optional field with a value of "2" for the left and
+    right channels.
 
-receive-data-capability = {
-  format: text ; we define "text-cue" for text cues
-  format-parameters: codec-parameters
-}
+: min-bit-rate (optional)
```
pthatcherg: Needs  a unit.  kbps is probably good.   And usually uint is good here.  No one is precise enough in bitrates to care about floats.

index.bs
```diff
@@ -2375,12 +2402,64 @@ data-frame = {
...
+receive-audio-capability = {
+  0: codec ; codec
+  ? 1: uint ; max-audio-channels
+  ? 2: float64 ; min-bit-rate
+}
+
+resolution = {
+  0: uint ; horizontal-resolution
+  1: uint ; vertical-resolution
+}
```
pthatcherg: I'd prefer:

height 
width

That's what it's called everywhere I've seen it

index.bs
```diff
@@ -1224,52 +1224,80 @@ as the *sender* and the agent receiving the media as the *receiver*.
...
+:: An optional field composed of a resolution, which includes (1) an unsigned
+    integer indicating horizontal resolution, and (2) an unsigned integer
+    indicating vertical resolution. The device can pass this field to indicate
+    the maximum resolution it supports.
+
+: max-frames-per-second (optional)
+:: An optional field indicating the maximum frames-per-second the device is
+    capable of processing.
+
+: min-video-bit-rate (optional)
```
pthatcherg: Units again

index.bs
```diff
@@ -1224,52 +1224,80 @@ as the *sender* and the agent receiving the media as the *receiver*.
...
+additional fields:
 
-receive-video-capability = {
-  codec-name: text
-  codec-parameters: [* codec-parameter]
-  ? max-resolution: video-resolution
-  ? max-frames-per-second: float
-  ? max-pixels-per-second: float
-}
+: max-audio-channels (optional)
```
pthatcherg: Instead of "device" should probably say "receiving agent" or "agent" or "receiver" here.  It might not be a "device".  I think I'd prefer "receiver".  Here and below.

index.bs
```diff
@@ -1224,52 +1224,80 @@ as the *sender* and the agent receiving the media as the *receiver*.
...
+: min-video-bit-rate (optional)
+:: An optional field indicating the minimum video bit rate the device is
+    capable of processing.
+
+: aspect-ratio (optional)
+:: An optional field indicating what its ideal aspect ratio is, e.g. a 16:10
+    display could return this value as 1.6 to indicate its preferred content
+    scaling.
+
+: color-profiles (optional)
```
pthatcherg: We need some examples of values with exact strings and probably somewhere to point where they are defined.

index.bs
```diff
@@ -1224,52 +1224,80 @@ as the *sender* and the agent receiving the media as the *receiver*.
...
+
+: max-frames-per-second (optional)
+:: An optional field indicating the maximum frames-per-second the device is
+    capable of processing.
+
+: min-video-bit-rate (optional)
+:: An optional field indicating the minimum video bit rate the device is
+    capable of processing.
+
+: aspect-ratio (optional)
```
pthatcherg: Hmm.... I hate floats.  I wonder if we should define a ratio type like we do for media time and then use that for aspect ratio and framerate.

index.bs
pthatcherg: Default is none?  If so, it's not optional because you can't send it anything!

index.bs
```diff
@@ -1224,52 +1224,80 @@ as the *sender* and the agent receiving the media as the *receiver*.
...
+    scaling.
+
+: color-profiles (optional)
+:: An optional field indicating what color profiles are understood, such as
+    sRGB v4. The listener may use these values to determine how to encode
+    video.
+
+: native-resolutions (optional)
+:: An optional field indicating what resolutions the device supports and
+    considers to be "native." If provided, the supports-scaling bool should
```
pthatcherg: "native" meaning scaling won't be required?  We should probably clarify that.  In fact, we might consider renaming it to "resolutions-that-dont-require-scaling" or some shorter version of that.

index.bs
```diff
@@ -1224,52 +1224,80 @@ as the *sender* and the agent receiving the media as the *receiver*.
...
+:: An optional field indicating what its ideal aspect ratio is, e.g. a 16:10
+    display could return this value as 1.6 to indicate its preferred content
+    scaling.
+
+: color-profiles (optional)
+:: An optional field indicating what color profiles are understood, such as
+    sRGB v4. The listener may use these values to determine how to encode
+    video.
+
+: native-resolutions (optional)
```
pthatcherg: Default is none?

index.bs
```diff
@@ -1224,52 +1224,80 @@ as the *sender* and the agent receiving the media as the *receiver*.
...
+:: An optional field indicating what resolutions the device supports and
+    considers to be "native." If provided, the supports-scaling bool should
+    be passed if the device is capable of scaling resolutions not in this list.
+
+: supports-scaling (optional)
+:: If omitted, this value is assumed to be false. Setting to "True" indicates
```
pthatcherg: Just say "the default is false" and move that to the end of the paragraph.

pthatcherg: Also, having the native resolutions and scaling be optional means if they are both omitted, you can't send anything, so they aren't really optional.  Perhaps we should have this default to true?

index.bs
```diff
@@ -2375,12 +2402,64 @@ data-frame = {
...
+resolution = {
+  0: uint ; horizontal-resolution
+  1: uint ; vertical-resolution
+}
+
+receive-video-capability = {
+  0: codec ; codec
+  ? 1: resolution ; max-resolution
+  ? 2: float64 ; max-frames-per-second
+  ? 3: float64 ; max-pixels-per-second
```
pthatcherg: This was never defined.

index.bs
```diff
@@ -2375,12 +2402,64 @@ data-frame = {
...
+  1: bytes ; value
+}
+
+receive-audio-capability = {
+  0: codec ; codec
+  ? 1: uint ; max-audio-channels
+  ? 2: float64 ; min-bit-rate
+}
+
+resolution = {
```
pthatcherg: Call this video-resolution?

baylesj: Sounds good.

index.bs
baylesj: Defined. Maybe there's a better definition I could do? Hm.

index.bs
baylesj: Default true sounds great to me.

index.bs
baylesj: Sounds right to me, assuming supports-scaling is default true. Updated.

index.bs
baylesj: I'll clarify the comment. FWIW, my understanding is that LCD screens will only have one native resolution, and all others require scaling. I see the usage of this in the old spec, I would be interested to see when we might want to pass multiple native resolutions--maybe the hardware requirements of the encoder?

index.bs
baylesj: Sounds good to me, I went back and forth on this.

index.bs
baylesj: agreed.

index.bs
baylesj: +1

index.bs
baylesj: Agreed. I went back and forth internally, this sounds great.

index.bs
baylesj: +1

index.bs
baylesj: +1

index.bs
baylesj: 2

index.bs
baylesj: yes

index.bs
baylesj: Done

index.bs
baylesj: Done.

index.bs
baylesj: Sounds good.

index.bs
baylesj: Switching to all "receiver" verbage.

index.bs
baylesj: Fair enough, lol. Probably sRGBv4?

index.bs
baylesj: Yeah, floats are dumb. Defined a ratio, are the words antecedent and consequent too bookish? It's what the terms of the ratio are called, but we could do something like "top" and "bottom" if that's too much.

index.bs
baylesj: I haven't found a good reference. Is there someone who would be a good resource for knowing if we actually need this? I see it in the old specs, and I know that there are oodles of color profiles, but it's not clear how much we actually need this.

index.bs
baylesj: Yes. We could default to 1080P or something, though I'm not sure if that has much value. I definitely want to get more domain knowledge in this area, I know a decent amount about video/media but definitely not sure about some of these decisions.

PR #155, opened 2019-04-05
Fix Typos in CDDL Spec

rwkeane:


messages_appendix.cddl
```diff
@@ -27,8 +27,10 @@ agent-status-response = {
   ? 1: status ; status
 }
 
-status = {
-}
+status = &(
+  success: 1
+  failure: 2
+)
```
pthatcherg: The empty struct was intentional.  It's the place where people can add things ad-hoc or non-standard.

rwkeane: Fixed

PR #156, opened 2019-04-08
Add explanation of integer key usage

baylesj:
Currently, we consistently use integer keys in our object
definitions, with the human readable field name appended as a comment.
This unusual syntax has significant memory benefits for us, but could
use an explanation for people not familiar with this design.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1514,7 +1514,12 @@ Appendix A: Messages {#appendix-a}
 
 The following messages are defined with [[CDDL]]. When
 integer keys are used, a comment is appended to the line to indicate
-the name of the field. Each root message (one that can be put into a
+the name of the field. This unusual syntax is consistently applied for object
+definitions in this specification due to it resulting in significantly smaller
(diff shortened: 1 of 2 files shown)


baylesj:
@mfoltzgoogle can you review/merge?

index.bs
```diff
@@ -1514,7 +1514,12 @@ Appendix A: Messages {#appendix-a}
 
 The following messages are defined with [[CDDL]]. When
 integer keys are used, a comment is appended to the line to indicate
-the name of the field. Each root message (one that can be put into a
+the name of the field. This unusual syntax is consistently applied for object
+definitions in this specification due to it resulting in significantly smaller
```
mfoltzgoogle: Suggested phrasing:

> Object definitions in this specification have this unusual syntax to reduce the number of bytes-on-the-wire, while maintaining a human-readable name for each key.

baylesj: Sounds good to me.

baylesj:
@mfoltzgoogle, should be good to go!

PR #157, opened 2019-04-09
Update authentication CDDL message definitions

maxyakimakha:


messages_appendix.cddl
```diff
@@ -41,18 +41,17 @@ response = (
 request-id = uint
 
 ; type key 1001
-authentication-request = {
+auth-request-hkdf-scrypt-psk = {
```
pthatcherg: We need to update the index.bs file at the same time to reference the new names and changes.

amills345: > We need to update the index.bs file at the same time to reference the new names and changes.

mfoltzgoogle:
I updated the spec prose in index.bs to reflect the new definitions and merged this PR.  We can review the changes at the F2F meeting.

PR #172, opened 2019-05-23
Finish media streaming (session start, stats)

pthatcherg:
Also remove streaming data, since it doesn't really have a use case

***
<a href="https://pr-preview.s3.amazonaws.com/webscreens/openscreenprotocol/pull/172.html" title="Last updated on Sep 10, 2019, 5:08 PM UTC (96dca50)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webscreens/openscreenprotocol/172/fe57b32...96dca50.html" title="Last updated on Sep 10, 2019, 5:08 PM UTC (96dca50)">Diff</a>


mfoltzgoogle:
@pthatcherg I rebased this PR which required several manual fixes.  Please take a look and make sure that nothing looks awry. 

I'll start a review shortly.

index.bs
```diff
@@ -1432,8 +1429,131 @@ following additional fields:
 Sessions {#streaming-sessions}
 ------------------------------------
 
-TODO
+To start a streaming session, a sender may send a
```
mfoltzgoogle: What if the receiver wants to start a streaming session?

mfoltzgoogle: See, e.g., Issue #176.

index.bs
```diff
@@ -1432,8 +1429,131 @@ following additional fields:
...
+: media-stream-id
+:: Identifies the media stream b	eing offered.  Must be unique within
+    the streaming session.  Can be used by the receiver to request the
+    media session.
+
+: friendly-name
+:: An optional name intended to be shown to a user, such that the
+    receiver may allow the user to choose which media streams to
+    receive, or if they are received automatically by the receiver,
+    give the user some information about what th e media stream is.
```
mfoltzgoogle: Typo in "the"

index.bs
```diff
@@ -1432,8 +1429,131 @@ following additional fields:
...
+:: Indicates the streams that the receiver can request from the sender.
+
+Each stream offer contains the following properties:
+
+: media-stream-id
+:: Identifies the media stream b	eing offered.  Must be unique within
+    the streaming session.  Can be used by the receiver to request the
+    media session.
+
+: friendly-name
```
mfoltzgoogle: Nit: display-name, for consistency with agent-info.display-name

index.bs
```diff
@@ -1432,8 +1429,131 @@ following additional fields:
...
+    sample rate.  They can differ by codec and related properties,
+    but should be different encodings of the same audio.
+
+: video
+:: A list of video encodings offered.  A video encoding is a series of
+    encoded video frames.  Encodings define properties needed by the
+    receiver to know how to decode the encoding, such as codec and
+    default duration.  They can differ by codec and potentially other
+    properties, but should be different encodings of the same video.
+
```
mfoltzgoogle: List of data encodings?

index.bs
```diff
@@ -1432,8 +1429,131 @@ following additional fields:
...
+
+Each audio encoding offered defines the following properties:
+
+: encoding-id
+:: Identifies the audio encoding being offered.  Must be unique within
+    the media stream.  Can be the receiver to request an encoding.
```
mfoltzgoogle: s/media stream/media streaming session/?

mfoltzgoogle: "Can be the receiver to request an encoding" is a sentence fragment

index.bs
```diff
@@ -1432,8 +1429,131 @@ following additional fields:
...
+: codec
+:: The name of the codec used by the encoding.
+
+: time-scale
+:: The time scale used by all audio frames.  This allows senders to
+    make audio-frame messages smaller by not including the time scale
+    in each one.
+
+: default-duration:
+:: The duration of an audio frame .  This allows senders to make
```
mfoltzgoogle: Extra space before period

index.bs
```diff
@@ -1432,8 +1429,131 @@ following additional fields:
...
+    audio-frame messagse smaller by not including the duration for
+    audio-frame messages that have the default duration.
+
+Each video encoding offered defines the following properties:
+
+: encoding-id
+:: Identifies the video encoding being offered.  Must be unique within
+    the media stream.  Can be the receiver to request an encoding.
```
mfoltzgoogle: Similar comments as above for encoding-id

index.bs
```diff
@@ -1432,8 +1429,131 @@ following additional fields:
...
+
+
+Each audio encoding offered defines the following properties:
+
+: encoding-id
+:: Identifies the audio encoding being offered.  Must be unique within
+    the media stream.  Can be the receiver to request an encoding.
+
+: codec
+:: The name of the codec used by the encoding.
```
mfoltzgoogle: This should probably refer to the RFCs that talk about codec strings used in MIME.
I think there is an open task to look at this from the F2F, so fine to address in a separate PR.

index.bs
```diff
@@ -1432,8 +1429,131 @@ following additional fields:
...
+
+: time-scale
+:: The time scale used by all video frames.  This allows senders to
+    make video-frame messages smaller by not including the time scale
+    in each one.
+
+: default-duration:
+:: The default duration of a video frame .  This allows senders to make
```
mfoltzgoogle: Extra space before period

index.bs
```diff
@@ -1432,8 +1429,131 @@ following additional fields:
...
+    video-frame messagse smaller by not including the duration for
+    video-frame messages that have the default duration.
+
+: default-rotation:
+:: The default rotation of a video frame .  This allows senders to make
```
mfoltzgoogle: Ditto

index.bs
```diff
@@ -1432,8 +1429,131 @@ following additional fields:
...
-TODO
+To start a streaming session, a sender may send a
+streaming-session-start-request message with the following properties:
 
+: streaming-session-id
+:: Identifies the streaming session.  Must be unique for the (sender,
+    receiver) pair.  Can be used later to modify or terminate a
+    streaming session.
+
+: stats-interval
```
mfoltzgoogle: Is the receiver required to support arbitrary stats intervals?

Maybe this should be a "desired" value, and the receiver is allowed to send stats at most that often, or less often if that's not possible.

index.bs
```diff
@@ -1432,8 +1429,131 @@ following additional fields:
...
+:: The default rotation of a video frame .  This allows senders to make
+    video-frame messagse smaller by not including the rotation for
+    video-frame messages that have the default rotation.
+
+After receiving a streaming-session-start-request message, a receiver
+should send back a streaming-session-start-response message with the
+following properties:
+
+: stats-interval
+:: Indicates the frequency the sender should send stats messages to
```
mfoltzgoogle: Similar comment as above regarding desired vs. actual stats intervals.

index.bs
```diff
@@ -1432,8 +1429,131 @@ following additional fields:
...
+    the receiver.
+
+: stream-requests
```
mfoltzgoogle: Does an empty stream-requests mean that the receiver can't support any of the possible encodings?

index.bs
```diff
@@ -1432,8 +1429,131 @@ following additional fields:
...
+:: Indicates which media streams the receiver would like to receiver
+    from the sender.
+
+Each stream request contains the following properties:
+
+: media-stream-id
+:: The ID of the stream reqeusted.
+
+: audio (optional)
+:: The requested audio encoding, by encoding ID
```
mfoltzgoogle: Is the receiver allowed to ask for subsampled audio, because of hardware limitations?  For example, 16 kHz instead of an offered encoding of 48 Khz.

Or, maybe the sender can offer multiple audio encodings at reasonable subsampling rates, if it is willing to subsample.

index.bs
```diff
@@ -1648,6 +1829,12 @@ Note: Request and watch IDs are not unique across agents.  An agent can combine
 a request ID with a unique identifier for the agent that sent it (like its
 certificate fingerprint) to track requests across multiple agents.
 
+Extension fields should use string keys to avoid conflicts with integer keys in
```
mfoltzgoogle: I think this was fallout from the merging/rebasing.  It should go at the end of Protocol Extensions.  I can fix this in a commit.

index.bs
```diff
@@ -1432,8 +1429,131 @@ following additional fields:
...
+:: The requested audio encoding, by encoding ID
+
+: video (optional)
+:: The requested video encoding, by encoding ID.  It may
+    include a target resolution and maximum frame rate.  The sender
+    should not exceed the maximum frame rate and should attempt to
+    send at the target bitrate, possibly exceeding it by a small amount.
+
+
+During a streaming session, the receiver can modify the requests it
```
mfoltzgoogle: I am assuming this allows the receiver to request a new (audio,video) encoding pair, that was already offered by the sender in 

What if the sender wants to change the encodings it offered?  Should it tear down the current session and start anew?

index.bs
```diff
@@ -1432,8 +1429,131 @@ following additional fields:
...
+:: The name of the codec used by the encoding.
+
+: time-scale
+:: The time scale used by all video frames.  This allows senders to
+    make video-frame messages smaller by not including the time scale
+    in each one.
+
+: default-duration:
+:: The default duration of a video frame .  This allows senders to make
+    video-frame messagse smaller by not including the duration for
```
mfoltzgoogle: Typo in message

index.bs
```diff
@@ -1432,8 +1429,131 @@ following additional fields:
...
+:: The name of the codec used by the encoding.
+
+: time-scale
+:: The time scale used by all audio frames.  This allows senders to
+    make audio-frame messages smaller by not including the time scale
+    in each one.
+
+: default-duration:
+:: The duration of an audio frame .  This allows senders to make
+    audio-frame messagse smaller by not including the duration for
```
mfoltzgoogle: Typo in messages

index.bs
```diff
@@ -1432,8 +1429,131 @@ following additional fields:
...
+    send at the target bitrate, possibly exceeding it by a small amount.
+
+
+During a streaming session, the receiver can modify the requests it
+made for encodings by sending a streaming-session-modify-request
+containing a modify list of stream-requests.  When the sender receives
+a streaming-session-modify-request, it should send back a
+streaming-session-modify-response.
+
+Finally, either sender may terminate the streaming session by sending
```
mfoltzgoogle: either agent ?

index.bs
```diff
@@ -1432,8 +1429,131 @@ following additional fields:
...
+    include a target resolution and maximum frame rate.  The sender
+    should not exceed the maximum frame rate and should attempt to
+    send at the target bitrate, possibly exceeding it by a small amount.
+
+
+During a streaming session, the receiver can modify the requests it
+made for encodings by sending a streaming-session-modify-request
+containing a modify list of stream-requests.  When the sender receives
+a streaming-session-modify-request, it should send back a
+streaming-session-modify-response.
```
mfoltzgoogle: How should the sender populate the values in the response?

index.bs
```diff
@@ -1619,7 +1693,114 @@ QUIC streams.
...
-TODO
+Durings a streaming session, the sender should send stats with the
+streaming-session-sender-stats-event at the interval the receiver
+requested.  The streaming-session-sender-stats-event message contains
+the following properties:
+
+: streaming-session-id
+:: The ID of the streaming session these stats apply to.
+
+: system-time
```
mfoltzgoogle: How would the receiver correlate these stats with specific audio or video frames?
Maybe the more relevant timing information is a min-max start time for the corresponding frames.

mfoltzgoogle: On the other hand, there may be intervals where no frames are being sent.  So maybe there needs to be a duration field, covering the wall clock time that elapsed during the stats snapshot.

index.bs
```diff
@@ -1619,7 +1693,114 @@ QUIC streams.
...
+:: Stats specific to video.  Stats for multiple encodings can be sent
+    at once, but encodings need not be included if the stats haven't
+    changed.  See below.
+
+Audio encoding sender stats include the following properties:
+
+: encoding-id
+:: The ID of the encoding for which the stats apply.
+
+: cumulative-sent-frames
```
mfoltzgoogle: Is there an advantage to cumulative vs. incremental metrics for these stats?

index.bs
```diff
@@ -1619,7 +1693,114 @@ QUIC streams.
...
+
+: cumulative-encode-delay
+:: The sum of the time spent encoding frames sent.
+
+: cumulative-dropped frames
+:: The total number of frames that were not sent due to network, CPU,
+    or other contraints.
+
+
+Durings a streaming session, the receiver should send stats with the
```
mfoltzgoogle: Typo in during

index.bs
```diff
@@ -1619,7 +1693,114 @@ QUIC streams.
...
+streaming-session-receiver-stats-event at the interval the sender
+requested.  The streaming-session-receiver-stats-event message contains
+the following properties:
+
+: streaming-session-id
+:: The ID of the streaming session these stats apply to.
+
+: system-time
```
mfoltzgoogle: Similar comments about frame range + duration.

index.bs
```diff
@@ -1619,7 +1693,114 @@ QUIC streams.
...
+: encoding-id
+:: The ID of the encoding for which the stats apply.
+
+: cumulative-decoded-frames
+:: The total number of video frames received and decoded.
+
+: cumulative-lost-frames
+:: The total number of video frames detected as lost.
+
+: cumulative-jitter-buffer-delay
```
mfoltzgoogle: I don't see any reference to a jitter buffer in the spec...

messages_appendix.cddl
```diff
@@ -549,7 +505,159 @@ receive-video-capability = {
...
 }
 
-receive-data-capability = {
-  0: format ; format
+; type key 134
+streaming-session-start-request = {
+  request
+  1: uint; streaming-session-id
+  2: [* media-stream-offer] ;  stream-offers
+  3: microseconds ; stats-interval
```
mfoltzgoogle: Milliseconds seems like it would be sufficient, but fine to keep microseconds for consistency with watch durations.

messages_appendix.cddl
```diff
@@ -549,7 +505,159 @@ receive-video-capability = {
...
+  1 : uint ; streaming-session-id
+}
+
+; type key 139
+streaming-session-terminate-response = {
+  response
+}
+
+; type key 140
+streaming-session-terminate-event = {
```
mfoltzgoogle: I didn't see this message referenced in the spec.

messages_appendix.cddl
```diff
@@ -549,7 +505,159 @@ receive-video-capability = {
...
+  antecedent: uint
+  consequent: uint
+]
+
+
+; type key 142
+streaming-session-sender-stats-event = {
+  1: uint; streaming-session-id
+  2: microseconds ; system-time
+  ? 3: [1* {
```
mfoltzgoogle: Why is this dictionary nested in an array (versus just being a nested dictionary)? 
Same comment applies to other arrays here.

messages_appendix.cddl
```diff
@@ -549,7 +505,159 @@ receive-video-capability = {
...
+}
+
+; type key 141
+video-frame-request = {
+ 1: uint; encoding-id
+ 2: uint; sequence-number
+ ? 3: uint; highest-decoded-frame-sequence-number
+}
+
+ratio = [
```
mfoltzgoogle: Is this used?

messages_appendix.cddl
```diff
@@ -549,7 +505,159 @@ receive-video-capability = {
...
+}
+
+audio-encoding-request = {
+  1: uint ; encoding-id
+}
+
+video-encoding-request = {
+  1 : uint ; encoding-id
+  ? 2 : video-resolution ; target-resolution
+  ? 3 : float32 ; max-frames-per-second
```
mfoltzgoogle: Will this ever be a non-integer?

messages_appendix.cddl
```diff
@@ -549,7 +505,159 @@ receive-video-capability = {
...
+  3 : uint ; time-scale
+  ? 4 : uint ; default-duration
+}
+
+video-encoding-offer = {
+  1 : uint ; encoding-id
+  2 : text ; codec
+  3 : uint ; time-scale
+  ? 4 : uint ; default-duration
+  ? 5 : video-rotation ; default-rotation
```
mfoltzgoogle: Is video-rotation defined?

messages_appendix.cddl
```diff
@@ -549,7 +505,159 @@ receive-video-capability = {
...
+
+; type key 14
+video-frame = {
+  0: uint ; encoding-id
+  1: uint ; sequence-number
+  ? 2: [* int] ; depends-on
+  3: uint ; start-time
+  ? 4: uint ; duration
+  5: bytes ; payload
+  ? 6: uint ; rotation
```
mfoltzgoogle: What are legal values?  Or should this be video-rotation?

messages_appendix.cddl
```diff
@@ -549,7 +505,159 @@ receive-video-capability = {
...
+    ? 1: media-time ; sync-time
+  }
+]
+
+; type key 14
+video-frame = {
+  0: uint ; encoding-id
+  1: uint ; sequence-number
+  ? 2: [* int] ; depends-on
+  3: uint ; start-time
```
mfoltzgoogle: The spec says that start-time and duration for an offer are inferred from the properties of the offer.  But it doesn't say exactly how.

messages_appendix.cddl
```diff
@@ -549,7 +505,159 @@ receive-video-capability = {
...
+]
+
+
+; type key 142
+streaming-session-sender-stats-event = {
+  1: uint; streaming-session-id
+  2: microseconds ; system-time
+  ? 3: [1* {
+    1: uint ; encoding-id
+    ? 2: uint ; cumulative-sent-frames
```
mfoltzgoogle: All the metrics are defined as optional fields.  But the spec doesn't go into details about when they should be set.  I'm assuming this is a way of delta encoding the stats, but some specific explanation of that would be good.

index.bs
mfoltzgoogle: Fixed

index.bs
```diff
@@ -1546,52 +1666,6 @@ ending at the last dependent frame.
     inferred from the properties of the encoding.
 
 
-Data {#media-streaming-data}
```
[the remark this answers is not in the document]
pthatcherg: OK, I added it back.  But I removed the part about text tracks because I think we should instead make a first-class message for text tracks.

index.bs
```diff
@@ -1432,8 +1429,131 @@ following additional fields:
...
+:: Indicates the frequency the receiver should send stats messages to
+    the sender.
+
+: stream-offers
+:: Indicates the streams that the receiver can request from the sender.
+
+Each stream offer contains the following properties:
+
+: media-stream-id
+:: Identifies the media stream b	eing offered.  Must be unique within
```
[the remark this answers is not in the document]
pthatcherg: done

index.bs
```diff
@@ -1432,8 +1429,131 @@ following additional fields:
...
+    the streaming session.  Can be used by the receiver to request the
```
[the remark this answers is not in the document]
pthatcherg: Yes, but that sounds like an implementation detail that we don't need to specify.

index.bs
pthatcherg: The receiver wants to start sending?  It could take on the role of media sender and start a session in the other direction.

The receiver wants to start receiving?  It could send a "please start a streaming session with me" message, I guess.   We could do something more complex to save a round trip, but I don't think that's worth it (SDP offer/answer has way too much complexity to do this 1-round-trip offer-to-receive type stuff).  But assuming we add such a message "streaming-session-receive-request", what would the receiver say it wants to receive?  "Please send me your desktop?"  "Please send me a tab"?  What would the controller do?  Pop up a some UI to the user saying "a TV wants you to stream it something"?  We don't have anything like this in the Presentation and Remote Playback APIs, do we?

index.bs
```diff
@@ -1432,8 +1429,131 @@ following additional fields:
 Sessions {#streaming-sessions}
 ------------------------------------
 
-TODO
+To start a streaming session, a sender may send a
+streaming-session-start-request message with the following properties:
 
+: streaming-session-id
+:: Identifies the streaming session.  Must be unique for the (sender,
```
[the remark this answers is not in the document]
pthatcherg: The receiver only knows about the IDs that the sender has requested and the receiver has responded to.  The receiver, when demuxing messages, just needs to demux by streaming-session-id.  It doesn't need to be unique across the different senders the receivers receives from (and thus doesn't need a sender-specific prefix).

index.bs
pthatcherg: done

index.bs
pthatcherg: Re-added.

index.bs
pthatcherg: Similar to above, as long as it's unique to the media stream, the receiver can demux based on it and the sender knows all the IDs the receiver knows because it picked them.

index.bs
pthatcherg: Changed to "desired".

index.bs
pthatcherg: Done here too

index.bs
pthatcherg: It's not unique to the session, but the particular media stream (there can be several streams).   removed the sentence fragment.

index.bs
pthatcherg: The later, but I believe Opus is always 48khz.   Downsampling before playout on the receiver isn't a big deal.

messages_appendix.cddl
pthatcherg: I just added it :).

messages_appendix.cddl
pthatcherg: video-rotation.

Fixed!

messages_appendix.cddl
pthatcherg: Time-scale and duration may be if if omitted.  But not start-time, and not if omitted.

messages_appendix.cddl
pthatcherg: Yes, it can be something like 30000/1001 (NTSC video).  So maybe we should make it a ratio.

messages_appendix.cddl
pthatcherg: I think we should say that all fields should be sent.  But if we make them required, that will be a pain in the future when we want to change stats around (which will invariably happen).

index.bs
pthatcherg: I'll try and do that in a separate PR.

pthatcherg:
This is ready for another look.

index.bs
pthatcherg: fixed

index.bs
pthatcherg: done

index.bs
pthatcherg: Yes, that would require a new session, at least currently.  Do we have a use case where a sender wants to offer new encodings but wouldn't want to restart a session?

index.bs
pthatcherg: Should just be "the sender".  Fixed.

index.bs
pthatcherg: It's basically just an OK/error response.

index.bs
pthatcherg: All the stats are meant to be "duration less" by using cumulative stats such that any two points in time can be compared to compute changes over time.  That way if you lose some samples, you can still have meaningful data, and if you have all the samples, you can use what window you want for averages and other calculations.

I don't think these are meant to be correlated with individual frames, but I believe the system-time should line up with the sync-time on frames.

index.bs
pthatcherg: Yes, I mentioned them in the previous comment.  It's how WebRTC does stats as well.  It's a good technique, I think.

index.bs
pthatcherg: fixed

index.bs
pthatcherg: Same answer :).

index.bs
pthatcherg: I made more specific text that doesn't mention jitter buffers.

messages_appendix.cddl
pthatcherg: It's in this paragraph: "Finally, the sender may terminate the streaming session by sending
a streaming-session-terminate-request command.  When the receiver
receives the streaming-session-terminate-request, it should send back
a streaming-session-terminate-response.  The receiver can terminate at
any point and notify the sender by sending a
streaming-session-terminate-event message."

messages_appendix.cddl
pthatcherg: Because there can be many: one for each encoding ID.

index.bs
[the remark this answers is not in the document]
mfoltzgoogle: SGTM.

index.bs
mfoltzgoogle: This would be for allowing apps to request media streams from capture devices or remote renderers, for example from IP webcams, or maybe VR headsets.  Or an application on the receiver could generate a stream.

Maybe discuss later as part of the bidirectional case?  (This is a specific case of that.)

index.bs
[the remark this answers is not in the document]
mfoltzgoogle: OK, then it could be from the same counter that we use for requests and watches, which is unique per-sender.  Should we just tell implementers to do that?

The receiver internally will need to add the sender fp or some other prefix to demux.

index.bs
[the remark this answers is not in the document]
mfoltzgoogle: But why would we specify IDs allocation for the other protocols and not here?  It seems like we should respect the state-token for streaming as well as presentation and remote playback.

index.bs
mfoltzgoogle: Still there

index.bs
mfoltzgoogle: Still there?

index.bs
mfoltzgoogle: Still there?

index.bs
```diff
@@ -1484,8 +1480,163 @@ following additional fields:
...
+After receiving a streaming-session-start-request message, a receiver
+should send back a streaming-session-start-response message with the
+following properties:
+
+: desired-stats-interval
+:: Indicates the frequency the sender should send stats messages to
+    the receiver.
+
+: stream-requests
+:: Indicates which media streams the receiver would like to receiver
```
mfoltzgoogle: Second receiver should be "receive"

index.bs
mfoltzgoogle: SGTM

index.bs
mfoltzgoogle: During remoting, the encoding can change for a given video element.  For example, ads may be encoded differently from the main content.

For mirroring, we may want to switch codecs according to network conditions and content (e.g., lossy vs. lossless).

index.bs
mfoltzgoogle: Can you add some text to clarify?

index.bs
mfoltzgoogle: That's a good design.  I would propose that if there's a need for frame level stats, they would be a different message that is tied to frame IDs and not system time.

index.bs
mfoltzgoogle: SGTM

messages_appendix.cddl
mfoltzgoogle: Got it.  After this lands, we can go through and link all the message definitions which helps them stand out.

messages_appendix.cddl
mfoltzgoogle: OK.  I think we still need to figure out if a missing stat means it wasn't captured, or was delta encoded (no change from prior sample).

index.bs
mfoltzgoogle: Fixed

index.bs

index.bs
mfoltzgoogle: Fixed

index.bs
mfoltzgoogle: Fixed

index.bs

index.bs

messages_appendix.cddl
mfoltzgoogle: Sorry I didn't quite follow that response, maybe restate it?

index.bs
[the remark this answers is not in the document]
pthatcherg: It makes sense that we need to do the same thing with forgetting IDs that we do for request IDs, so I'll recommend that.

index.bs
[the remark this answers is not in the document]
pthatcherg: Yeah, same here.  We should use the state-token for all IDs.

index.bs
pthatcherg: In that case, I think a "please offer me what streaming you have" message would make sense.  We can add it in a separate PR.

index.bs
pthatcherg: Hmmm... in that case we need to go back to the drawing board a bit and go with a sender-controlled codec where the receiver gives is capabilities and then the sender chooses.

index.bs
pthatcherg: Sure

messages_appendix.cddl
pthatcherg: Yes, in receive-video-capability.aspect-ratio.

messages_appendix.cddl
pthatcherg: Time-scale and duration may be inferred.  

start-time cannot be inferred.

Inference is basically just "if it isn't there, use the default", but I guess I could make that more explicit.

index.bs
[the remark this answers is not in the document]
pthatcherg: OK, I specified that.

index.bs
pthatcherg: OK, I specified that.

index.bs
pthatcherg: I included that one as well.

index.bs
pthatcherg: Let's land this PR and do that in a separate PR.

messages_appendix.cddl
pthatcherg: Made more explicit now.

pthatcherg:
OK, I think I got everything other than the two things we'll include in separate PRs:

1. Codec switching when remoting
2. Requesting someone send you a streaming session.

index.bs
```diff
@@ -1516,13 +1528,184 @@ following additional fields:
...
+    should not exceed the maximum frame rate and should attempt to
+    send at the target bitrate, possibly exceeding it by a small amount.
+
+: data (optional)
+:: The requested data encoding, by encoding ID
+
+
+During a streaming session, the receiver can modify the requests it
+made for encodings by sending a `streaming-session-modify-request`
+containing a modify list of stream-requests.  When the sender receives
```
mfoltzgoogle: Note to self: Fix typo before merging

index.bs
```diff
@@ -1610,51 +1797,51 @@ ending at the last dependent frame.
...
+Data {#streaming-data}
 ------------------------------------
 
 Media senders may send timed data to media receivers by sending [=data-frame=] messages (see
 [[#appendix-a]]) with the following keys and values.  A data frame message
-contains an arbitrary payload that can be synchronized with and video, such as
-text track data.  A series of data frames that share a data type and timeline
+contains an arbitrary payload that can be synchronized with audio and video.
+A series of data frames that share a data type and timeline
 form a data encoding.
```
mfoltzgoogle: Note to self: Fix line wrapping

PR #173, opened 2019-05-23
Add support for remoting.

pthatcherg:
***
<a href="https://pr-preview.s3.amazonaws.com/webscreens/openscreenprotocol/pull/173.html" title="Last updated on May 23, 2019, 2:43 AM UTC (76ecd22)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webscreens/openscreenprotocol/173/fb68050...76ecd22.html" title="Last updated on May 23, 2019, 2:43 AM UTC (76ecd22)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -888,9 +888,13 @@ the receiver decides to not attempt to load the resource at all.  If it chooses
 not to, it must respond with the appropriate failure result (such as timeout or
 invalid-url).  Additionally, the response must include the following:
 
-: state
+: state (optional)
 :: The initial state of the remote playback, as defined in
     [[#remote-playback-state-and-controls]].
+
+: remoting-capabilities (optional)
+:: What kind of remoting (using the streaming protocol) that the
+    receiver can receive.
 
 If the controller wishes to modify the state of the remote playback (for
 example, to pause, resume, skip, etc), it may send a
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -902,13 +906,29 @@ remote-playback-modify-request message with the following values:
 : controls
 :: Updated controls as defined in {#remote-playback-state-and-controls}
 
+: remoting-audio
```
baylesj: Can this be omitted if we only support video?

index.bs
```diff
@@ -902,13 +906,29 @@ remote-playback-modify-request message with the following values:
...
+:: Audio encodings that the controller can send to the receiver once they
+    are acked via the remote-playback-modify-response message.
+
+: remoting-video
```
baylesj: Can this be omitted if we only support audio?

index.bs
```diff
@@ -902,13 +906,29 @@ remote-playback-modify-request message with the following values:
...
+:: Videos encodings that the controller can send to the receiver once they
+    are acked via the remote-playback-modify-response message.
+
+: remoting-stats-interval
```
baylesj: Looks like from the CDDL below that this is optional?

index.bs
chrisn: Remote playback of audio streams is a use case we'd like to see supported. I hope there is implementer interest!

Issue #131, opened 2019-01-03
[Security] Mitigations for remote network attackers


mfoltzgoogle:
One increasingly common vector of attack is to take control of local network services from the public Internet via misconfigured routers [1].  We landed PR #96 with mitigations for SSDP, but these won't directly apply to an mDNS+QUIC world.  This issue tracks adding mitigations to the 1.0 spec to defend against remote attacks.

[1] https://www.theverge.com/2019/1/2/18165386/pewdiepie-chromecast-hack-tseries-google-chromecast-smart-tv

mfoltzgoogle:
Support added for a token as proof of discovery and required to initiate authentication.
If we desire other mitigation strategies, we can open additional issues with requirements/proposals.

PR #182, opened 2019-08-08
Adds an unguessable token to mDNS.

mfoltzgoogle:
This PR addresses Issue #131 (Mitigations for remote network attackers) by adding an unguessable token to the mDNS TXT record.

This token must be included in the first authentication message sent to/from the agent that advertises it, to prevent remote attackers from attempting authentication.

It also fixes a couple of typos in the authentication section and clarifies that SPAKE2 is mandatory for agents to implement.

***
<a href="https://pr-preview.s3.amazonaws.com/webscreens/openscreenprotocol/pull/182.html" title="Last updated on Aug 14, 2019, 10:55 PM UTC (9ef6db6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webscreens/openscreenprotocol/182/ec8c149...9ef6db6.html" title="Last updated on Aug 14, 2019, 10:55 PM UTC (9ef6db6)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -61,8 +61,6 @@ urlPrefix: https://w3c.github.io/remote-playback/#dfn-; type: dfn; spec: REMOTE-
     text: remote playback source
 urlPrefix: https://www.w3.org/TR/html51/single-page.html; type: dfn; spec: HTML51
     text: media element
-url: https://tools.ietf.org/html/rfc4648#section-4; type: dfn; spec: RFC4648; text: Base64
-url: https://tools.ietf.org/html/rfc8445#section-5.3; type: dfn; spec: RFC8445; text: ICE password
 </pre>
 
 <h2 class='no-num no-toc no-ref' id='status'>Status of this document</h2>
@@ -322,13 +320,15 @@ Issue: Include cross references to the specs for these hash functions.
     value.  This signals to the listening agent that it should connect to the
     advertising agent to discover updated metadata.
 
-The advertising agent may add an additional field to the TXT record:
+The advertising agent should add an additional field to the TXT record:
 
 : pw
-:: An alphanumeric, unguessable token consisting of a [=Base64=] encoded 128-bit
-    random value (24 characters in length).
+:: An alphanumeric, unguessable token consisting of characters from the set
+    `[A-Za-z0-9+/]`.
 
-Note: The requirements for `pw` are the same as for an [=ICE password=].
+Note: `pw` prevents off-LAN parties from attempting authentication; see
+[[#remote-active-mitigations]].  `pw` should have at least 32 bits of true
+entropy to make brute force attacks impractical.
 
 Issue: Add examples of sample mDNS records.
 
@@ -499,12 +499,12 @@ that it's easy for the user to input PSK on the device.  Supported PSK input met
 are numeric and scanning a QR-code.  Devices with non-zero PSK ease of input must
 support the numeric PSK input method.
 
-Any authentication method may require an `auth-initiation-password` before
+Any authentication method may require an `auth-initation-token` before
 showing a PSK to the user or requesting PSK input from the user.  If an
 advertising agent has the `pw` field in its mDNS TXT record, it must be used as
-the `auth-initiation-password` in the the first authentication message sent to
+the `auth-initation-token` in the the first authentication message sent to
 or from that agent.  Agents should discard any authentication message whose
-`auth-initiation-password` is set and does not match the `pw` provided by the
+`auth-initation-token` is set and does not match the `pw` provided by the
 advertising agent.
 
 TODO: Autolink advertising agent when PR #180 lands.
--- a/index.bs
+++ b/index.bs
@@ -526,15 +526,13 @@ that it's easy for the user to input PSK on the device.  Supported PSK input met
 are numeric and scanning a QR-code.  Devices with non-zero PSK ease of input must
 support the numeric PSK input method.
 
-Any authentication method may require an `auth-initation-token` before
-showing a PSK to the user or requesting PSK input from the user.  If an
-advertising agent has the `pw` field in its mDNS TXT record, it must be used as
-the `auth-initation-token` in the the first authentication message sent to
-or from that agent.  Agents should discard any authentication message whose
+Any authentication method may require an `auth-initation-token` before showing a
+PSK to the user or requesting PSK input from the user.  If an [=advertising
+agent=] has the `pw` field in its mDNS TXT record, it must be used as the
+`auth-initation-token` in the the first authentication message sent to or from
+that agent.  Agents should discard any authentication message whose
 `auth-initation-token` is set and does not match the `pw` provided by the
 advertising agent.
-
-TODO: Autolink advertising agent when PR #180 lands.
 
 Authentication with SPAKE2 {#authentication-with-spake2}
 --------------------------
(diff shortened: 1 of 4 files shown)


mfoltzgoogle:
PTAL @maxyakimakha and @pthatcherg

mfoltzgoogle:
For references, here's a link to the minutes from Berlin where this was discussed:

https://www.w3.org/2019/05/23-webscreens-minutes.html#x24

index.bs
```diff
@@ -320,8 +322,15 @@ Issue: Include cross references to the specs for these hash functions.
     value.  This signals to the listening agent that it should connect to the
     advertising agent to discover updated metadata.
 
-Issue: Add examples of sample mDNS records.
+The advertising agent may add an additional field to the TXT record:
+
+: pw
+:: An alphanumeric, unguessable token consisting of a [=Base64=] encoded 128-bit
+    random value (24 characters in length).
```
pthatcherg: Why not just say a random alphanumeric string?  Why specify it has to be Base64?

index.bs
```diff
@@ -320,8 +322,15 @@ Issue: Include cross references to the specs for these hash functions.
...
 
+Note: The requirements for `pw` are the same as for an [=ICE password=].
```
pthatcherg: Not sure what this buys us.  It doesn't seem worth having a dependency on ICE.

index.bs
```diff
@@ -320,8 +322,15 @@ Issue: Include cross references to the specs for these hash functions.
     value.  This signals to the listening agent that it should connect to the
     advertising agent to discover updated metadata.
 
-Issue: Add examples of sample mDNS records.
+The advertising agent may add an additional field to the TXT record:
+
+: pw
```
pthatcherg: Why so terse?  Why not "auth-token"?

index.bs
```diff
@@ -490,6 +499,19 @@ that it's easy for the user to input PSK on the device.  Supported PSK input met
 are numeric and scanning a QR-code.  Devices with non-zero PSK ease of input must
 support the numeric PSK input method.
 
+Any authentication method may require an `auth-initiation-password` before
```
pthatcherg: Why not "auth-initiation-token"?

messages_appendix.cddl
```diff
@@ -66,12 +66,18 @@ psk-input-method = &(
...
+)
+
 ; type key 1002
 auth-spake2-need-psk = {
+  auth-initiation-password
 }
 
 ; type key 1003
 auth-spake2-message = {
+  auth-initation-password
```
pthatcherg: Is it needed for this message?  It seems like it's just needed for auth-spake2-need-psk.

index.bs
```diff
@@ -320,8 +322,15 @@ Issue: Include cross references to the specs for these hash functions.
     value.  This signals to the listening agent that it should connect to the
     advertising agent to discover updated metadata.
 
-Issue: Add examples of sample mDNS records.
+The advertising agent may add an additional field to the TXT record:
```
pthatcherg: Perhaps make it a "should" instead of "may" and briefly explain why one should do it.

index.bs
mfoltzgoogle: This is in the TXT record and we need to conserve bytes.  One character keys are possible according to DNS-SD, but two seems reasonable in case we have similar keys.

https://tools.ietf.org/html/rfc6763#section-6.4

index.bs
mfoltzgoogle: It can't be any "random string".   It needs to be unguessable and have some minimum quantity of true entropy.  And it needs to be encoded efficiently to not waste bytes in the TXT.

I removed the base64 requirement though.  The specific encoding isn't important.

index.bs
mfoltzgoogle: Agreed. Removed reference to ICE and just recommend a minimum entropy.  We can discuss the exact value at the next F2F.

index.bs

messages_appendix.cddl
mfoltzgoogle: It's needed when the PSK presenter initiates authentication.

index.bs

index.bs
```diff
@@ -340,8 +340,17 @@ Issue: Include cross references to the specs for these hash functions.
     value.  This signals to the listening agent that it should connect to the
     advertising agent to discover updated metadata.
 
-Issue: Add examples of sample mDNS records.
+The advertising agent should add an additional field to the TXT record:
+
+: pw
+:: An alphanumeric, unguessable token consisting of characters from the set
+    `[A-Za-z0-9+/]`.
```
baylesj: Why allow forward slash '/' and plus sign '+'?

index.bs
```diff
@@ -340,8 +340,17 @@ Issue: Include cross references to the specs for these hash functions.
     value.  This signals to the listening agent that it should connect to the
     advertising agent to discover updated metadata.
 
-Issue: Add examples of sample mDNS records.
+The advertising agent should add an additional field to the TXT record:
```
baylesj: Nit: should = mandatory or optional? What happens if they don't?

index.bs
```diff
@@ -517,6 +526,19 @@ that it's easy for the user to input PSK on the device.  Supported PSK input met
...
 
+Any authentication method may require an `auth-initation-token` before
+showing a PSK to the user or requesting PSK input from the user.  If an
+advertising agent has the `pw` field in its mDNS TXT record, it must be used as
+the `auth-initation-token` in the the first authentication message sent to
+or from that agent.  Agents should discard any authentication message whose
+`auth-initation-token` is set and does not match the `pw` provided by the
+advertising agent.
+
+TODO: Autolink advertising agent when PR #180 lands.
```
baylesj: Nit: TODOs should have a bug or an alias.

index.bs
```diff
@@ -548,17 +570,17 @@ auth-spake2-message, auth-spake2-confirmation and auth-status.
 SPAKE2 describes in detail how auth-spake2-message and auth-spake2-confirmation
 are computed.
 
-If the PSK presenter wants to perform authentication, the PSK presenter
-starts the authentication process by presenting the PSK to the user and sending
-a auth-spake2-message message. When the PSK consumer receives
-the auth-spake2-message message, the PSK consumer prompts the user for the PSK
-input if it has not done so yet.
+If the PSK presenter wants to perform authentication, the PSK presenter starts
```
baylesj: %s/perform authentication/authenticate

index.bs
mfoltzgoogle: If an agent doesn't advertise the token, it will have a lower level of protection against outside parties attempting to authenticate to them (it will rely on a firewall/NAT to block incoming connections).

Maybe this makes sense in some closed environments or use cases.  I think there's a case to make it mandatory, but we'll need to consider future compatibility with cross-LAN authentication.

Will file an issue to consider the implications here.

index.bs
mfoltzgoogle: This is the same character set as Base64.  It allows agents to use Base64 (which is widely implemented) to encode a binary token.  But per earlier feedback we're not going to require Base64 in case agents want to make a different choice.

index.bs
mfoltzgoogle: Filed Issue #185.

index.bs
mfoltzgoogle: Resolved this TODO.

index.bs
mfoltzgoogle: Fixed in both places.

messages_appendix.cddl
```diff
@@ -67,12 +67,18 @@ psk-input-method = &(
   qr-code: 1
 )
 
+auth-initiation-token = (
+  ? 0 : text ; token
+)
+
 ; type key 1002
 auth-spake2-need-psk = {
+  auth-initiation-password
```
pthatcherg: I think here and below should be "-token" since you made it "-token" everywhere else.

index.bs
pthatcherg: Oh, right.  I forgot about that.

Well, it's called "-token" every where else now, so "pw" doesn't seem to fit.  Perhaps "tk"?  "at"?

messages_appendix.cddl
maxyakimakha: +1

mfoltzgoogle: I'll fix this in a follow-up commit.

mfoltzgoogle:
Sorry, I missed the comment about pw vs at for the mDNS key.  Let's use at for consistency with the CBOR protocol.

Issue #143, opened 2019-03-22
[Protocol] [Privacy] Notify endpoints when new connection is created


mfoltzgoogle:
The Security & Privacy considerations support notifying all connected enpdpoints when a new presentation connection is made to a presentation.  However the spec only requires a response to the endpoint that initiated the request.

Propose broadcasting an event to all endpoints post-connection to handle this requirement, so that an OSP user agents can notify their users.

Ref: https://github.com/webscreens/openscreenprotocol/pull/133#discussion_r247293475
Ref: https://w3c.github.io/presentation-api/#user-interface-guidelines

mfoltzgoogle:
Closed by PR #187.

PR #187, opened 2019-08-19
Adds presentation-change-event.

mfoltzgoogle:
This addresses Issue #143: Notify endpoints when new connection is created.

It adds a presentation-change-event message that is sent to all other connected presentation controllers when a presentation receiver accepts an incoming connection request.

Discussed here at the F2F: https://www.w3.org/2019/05/23-webscreens-minutes.html#x22

***
<a href="https://pr-preview.s3.amazonaws.com/webscreens/openscreenprotocol/pull/187.html" title="Last updated on Aug 21, 2019, 4:09 PM UTC (3ab315d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webscreens/openscreenprotocol/187/2164e92...3ab315d.html" title="Last updated on Aug 21, 2019, 4:09 PM UTC (3ab315d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -840,7 +840,7 @@ When [[PRESENTATION-API#closing-a-presentationconnection|section 6.5.5]] says
 "Start to signal to the destination browsing context the intention to close the
 corresponding PresentationConnection", the [=user agent=] may send a
 `presentation-connection-close-event` message to the user agent with the
-destination browsing context.
+destination browsing context and a `presentation-change-event` when required.
 
 When
 [[PRESENTATION-API#terminating-a-presentation-in-a-controlling-browsing-context|section
--- a/index.bs
+++ b/index.bs
@@ -747,7 +747,7 @@ following values:
     across connections, thus making message demuxing/routing easier).
 
 : connection-count
-:: The new total number of connections open to the presentation that received
+:: The new number of open connections to the presentation that received
     the incoming connection request.
 
 If the `presentation-connection-open-response` message indicates success, the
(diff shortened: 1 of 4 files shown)


mfoltzgoogle:
PTAL @chrisn @pthatcherg

index.bs
```diff
@@ -746,6 +746,22 @@ following values:
...
     across connections, thus making message demuxing/routing easier).
 
+: connection-count
+:: The new total number of connections open to the presentation that received
+    the incoming connection request.
+
+If the `presentation-connection-open-response` message indicates success, the
+receiver should also send a `presentation-change-event` to all other endpoints
+that have an active presentation connection to that presentation with the
+values:
```
pthatcherg: Shouldn't we also send one when a presentation is closed (when the number goes down?)

chrisn: Makes sense, for symmetry. So we'd want to add the count to the `presentation-termination-response` and `presentation-termination-event` messages.

chrisn: Should the name be `presentation-connection-change-event` for consistency with the other `presentation-connection-xxx` messages we have?

mfoltzgoogle: @pthatcherg Good point.  Added connection-count to presentation-connection-close-event, and how have the receiver send a change-event if it closes a connection.

@chrisn I went with presentation-change-event since we are announcing a change to a property of the presentation itself (the number of open connections), instead of a change to an individual connection.

mfoltzgoogle: Also, @chrisn, terminating a presentation must close all of its connections.  So I don't think we need to add a count to the termination messages.

chrisn: Ah, I was thinking of connection termination rather than presentation termination. So, you're right, the change looks good. Thanks!

PR #208, opened 2019-09-04
Add psk-min-bits-of-entropy

pthatcherg:
***
<a href="https://pr-preview.s3.amazonaws.com/webscreens/openscreenprotocol/pull/208.html" title="Last updated on Sep 4, 2019, 8:26 PM UTC (8922c46)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webscreens/openscreenprotocol/208/d75855b...8922c46.html" title="Last updated on Sep 4, 2019, 8:26 PM UTC (8922c46)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -537,8 +537,7 @@ agents may specify the minimum bits of entropy it requires for a PSK, in the
 range of 20 to 60 bits inclusive.  Agents must terminate any connection
 attempting to authenticate with a PSK with fewer than the minimum number of bits
 of entropy and must not attempt to authenticate with an agent that advertises a
-value outside the allowed range (less than 20 or more than 60 bits of
-entropy).
+value outside the allowed range.
 
 Authentication with SPAKE2 {#authentication-with-spake2}
 --------------------------
--- a/index.bs
+++ b/index.bs
@@ -534,10 +534,9 @@ advertising agent.
 
 In the `psk-min-bits-of-entropy` field of the [=auth-capabilities=] messsage,
 agents may specify the minimum bits of entropy it requires for a PSK, in the
-range of 20 to 60 bits inclusive.  Agents must terminate any connection
-attempting to authenticate with a PSK with fewer than the minimum number of bits
-of entropy and must not attempt to authenticate with an agent that advertises a
-value outside the allowed range.
+range of 20 to 60 bits inclusive, with a default of 20.  The PSK presenter must
+generate a PSK that has at least as many bits of entropy as it receives in this
+field, and at least as many bits of entropy as it sends in this field.
 
 Authentication with SPAKE2 {#authentication-with-spake2}
 --------------------------
--- a/index.bs
+++ b/index.bs
@@ -532,6 +532,14 @@ that agent.  Agents should discard any authentication message whose
 `auth-initation-token` is set and does not match the `at` provided by the
 advertising agent.
 
+In the `psk-min-bits-of-entropy` field of the [=auth-capabilities=] messsage,
+agents may specify the minimum bits of entropy it requires for a PSK, in the
+range of 20 to 60 bits inclusive.  Agents must terminate any connection
+attempting to authenticate with a PSK with fewer than the minimum number of bits
+of entropy and must not attempt to authenticate with an agent that advertises a
+value outside the allowed range (less than 20 or more than 60 bits of
+entropy).
+
 Authentication with SPAKE2 {#authentication-with-spake2}
 --------------------------
 
--- a/messages_appendix.cddl
+++ b/messages_appendix.cddl
@@ -63,6 +63,7 @@ microseconds = uint
 auth-capabilities = {
   0: uint ; psk-ease-of-input
   1: [* psk-input-method] ; psk-input-methods
+  2: uint ; psk-min-bits-of-entropy
 }
 
 psk-input-method = &(


index.bs
```diff
@@ -532,6 +532,14 @@ that agent.  Agents should discard any authentication message whose
 `auth-initation-token` is set and does not match the `at` provided by the
 advertising agent.
 
+In the `psk-min-bits-of-entropy` field of the [=auth-capabilities=] messsage,
+agents may specify the minimum bits of entropy it requires for a PSK, in the
+range of 20 to 60 bits inclusive.  Agents must terminate any connection
```
baylesj: Why cap at 60? Isn't it pretty common for VPN type things for PSKs to have 128+ bits? Is there any reason other than performance to limit larger keys?

index.bs
```diff
@@ -532,6 +532,14 @@ that agent.  Agents should discard any authentication message whose
...
+attempting to authenticate with a PSK with fewer than the minimum number of bits
+of entropy and must not attempt to authenticate with an agent that advertises a
+value outside the allowed range (less than 20 or more than 60 bits of
```
baylesj: Nit: the range is already specified, I'd vote for not doubly defining.

index.bs
pthatcherg: Since we're (likely?) using memory hard functions, I might not have enough memory to do more than 60.  So if you say your minimum is 60bits, we're stuck.  Plus, in this context, 60bits is already overkill.

index.bs
pthatcherg: Removed

index.bs
```diff
@@ -532,6 +532,13 @@ that agent.  Agents should discard any authentication message whose
 `auth-initation-token` is set and does not match the `at` provided by the
 advertising agent.
 
+In the `psk-min-bits-of-entropy` field of the [=auth-capabilities=] messsage,
+agents may specify the minimum bits of entropy it requires for a PSK, in the
+range of 20 to 60 bits inclusive.  Agents must terminate any connection
+attempting to authenticate with a PSK with fewer than the minimum number of bits
+of entropy and must not attempt to authenticate with an agent that advertises a
+value outside the allowed range.
```
mfoltzgoogle: The text doesn't spell out how the agents agree on the bit length of the PSK.  I'm assuming that the PSK presenter is required to generate a PSK of length max([my psk-min-bits-of-entry], [your psk-min-bits-of-entropy]).   I think the text should include this rule.

index.bs
```diff
@@ -532,6 +532,13 @@ that agent.  Agents should discard any authentication message whose
 `auth-initation-token` is set and does not match the `at` provided by the
 advertising agent.
 
+In the `psk-min-bits-of-entropy` field of the [=auth-capabilities=] messsage,
+agents may specify the minimum bits of entropy it requires for a PSK, in the
+range of 20 to 60 bits inclusive.  Agents must terminate any connection
+attempting to authenticate with a PSK with fewer than the minimum number of bits
+of entropy and must not attempt to authenticate with an agent that advertises a
```
mfoltzgoogle: How would we upgrade this protocol to support > 60 bit PSKs in the future?

index.bs
pthatcherg: OK, I added that.

pthatcherg: FYI, I also added a default value.

index.bs
pthatcherg: If anyone ever requires more 60 bits and older agents can't do that many bits, then they would necessarily be incompatible.  I suppose we could represent that with a new auth-capability and then a new message field, or maybe a new message type key with a new message capability.

mfoltzgoogle: Ok, should make sure that any such future extension is safe from downgrade attacks (similar to the existing protocol).

mfoltzgoogle:
Does this address the following item from the F2F:

F2F Action: Include analysis that downgrade of number of bits of entropy can be detected and auth fails.  

It mentions that the agent can close the connection if it gets a PSK that's too short.  Maybe just add a note in the security and privacy section mentioning that this protects against downgrades.

nigelcearnshaw:
Related: Do we assume that n bits of entropy is always conveyed in a PSK of n bits. Probably fair, but it does mean we are always exercising the number range 0-2^n -1 when strictly speaking we don't care where the entropy bits are in the wider number space. There are after all  something of the order of 2^240 elements available in a 25519 curve. If some presenter devices add non random bits to the PSK, perhaps because of a fixed salt value, or a brand mark, or the time of day - all non random - they would actually be potentially exercising a different random set of 2^n elements from within the full set of 2^240 odd.

I agree that a bit field length of less than n is definitely a downgrade attack, but a downgrade attack might be concealed if an honest presenter was augmenting the n bits of entropy with other non random fields and presenting a PSK of n random +more non random bits

Should we specify entropy n = PSK bit field length or remain silent, potentially undermining this test.
Maybe allowing larger total bit field lengths causes performance issues?

mfoltzgoogle:
- We should be clear in the security analysis/implementation notes that the PSK should have N bits of true entropy from a secure RNG where N was negotiated as described in the PR.

- I opened Issue #210 for translating the PSK into a numeric or QR code.  There may be a reason for the PSK presenter to add a non-random salt for presentation, but that won't affect the code actually used in the SPAKE2 handshake.

Issue #148, opened 2019-04-01
Make a required/default remote playback state table


pthatcherg:
The remote playback API control protocol text (currently in PR) is a bit verbose explaining what is required in the initial state and what the defaults are when they aren't required.  Having a table would be nice.

takumif:
For the [state](https://webscreens.github.io/openscreenprotocol/#issue-6098c204) fields, we've only said that `source` is required in a `remote-playback-start-response` message. Is `source` required because the request may contain multiple URLs and the receiver must choose one?

Are all the other fields in a `remote-playback-start-response` message optional? For example when we say "The default is empty (support for nothing) for the initial state in the `remote-playback-start-response` message" (for `supports`), do we mean that the message doesn't need to set it, and if not set we'd just assume it's empty?

pthatcherg:
I don't remember why it's required, but that sounds like a good reason :).

Yes, I believe the rest of the fields are optional.

mfoltzgoogle:
It looks like this was done in PR #217 by @takumif (Thank you!)  I looked through the table and nothing looks amiss.

I will clarify why `source` is required in the `remote-playback-start-response` message and close this out.

PR #217, opened 2019-09-10
Add a table for Remote Playback controls and states

takumif:
This addresses #148. AFAICT none of the fields is required to be in the initial state sent, but please let me know if I misunderstood this.

***
<a href="https://pr-preview.s3.amazonaws.com/takumif/openscreenprotocol/pull/217.html" title="Last updated on Sep 15, 2019, 9:49 PM UTC (d0d213b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webscreens/openscreenprotocol/217/dea1dce...takumif:d0d213b.html" title="Last updated on Sep 15, 2019, 9:49 PM UTC (d0d213b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1234,7 +1234,25 @@ intentionally mirror settable attributes and methods of the
     if a cue ID is invalid (removing an un-added ID or adding an ID twice, for example),
     the receiver may reject the text track change.
 
-Issue: Add a table for whether it's required and what the default is.
+<div class="note">
+<table>
+<tr><th>Field</th>                  <th>Default value</th>                              <th>Receiver support</th></tr>
+<tr><td>source</td>                 <td>`urls` in [=remote-playback-start-request=]</td><td>Required</td></tr>
+<tr><td>preload</td>                <td>Decided by the receiver</td>                    <td>Not required</td></tr>
+<tr><td>loop</td>                   <td>False</td>                                      <td>Required</td></tr>
+<tr><td>paused</td>                 <td>Decided by the receiver</td>                    <td>Required</td></tr>
+<tr><td>muted</td>                  <td>Decided by the receiver</td>                    <td>Required</td></tr>
+<tr><td>volume</td>                 <td>Decided by the receiver</td>                    <td>Required</td></tr>
+<tr><td>seek</td>                   <td>-</td>                                          <td>Required</td></tr>
+<tr><td>fast-seek</td>              <td>-</td>                                          <td>Required</td></tr>
+<tr><td>playback-rate</td>          <td>Decided by the receiver</td>                    <td>Not required</td></tr>
+<tr><td>poster</td>                 <td>Decided by the receiver</td>                    <td>Not required</td></tr>
+<tr><td>enabled-audio-track-ids</td><td>-</td>                                          <td>Required</td></tr>
+<tr><td>select-video-track-id</td>  <td>-</td>                                          <td>Required</td></tr>
+<tr><td>added-text-tracks</td>      <td>-</td>                                          <td>Not required</td></tr>
+<tr><td>changed-text-tracks</td>    <td>-</td>                                          <td>Not required</td></tr>
+</table>
+</div>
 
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -1234,7 +1231,25 @@ intentionally mirror settable attributes and methods of the
...
+<tr><td>loop</td>                   <td>False</td>                                      <td>Required</td></tr>
+<tr><td>paused</td>                 <td>Decided by the receiver</td>                    <td>Required</td></tr>
+<tr><td>muted</td>                  <td>Decided by the receiver</td>                    <td>Required</td></tr>
+<tr><td>volume</td>                 <td>Decided by the receiver</td>                    <td>Required</td></tr>
+<tr><td>seek</td>                   <td>-</td>                                          <td>Required</td></tr>
+<tr><td>fast-seek</td>              <td>-</td>                                          <td>Required</td></tr>
+<tr><td>playback-rate</td>          <td>Decided by the receiver</td>                    <td>Not required</td></tr>
+<tr><td>poster</td>                 <td>Decided by the receiver</td>                    <td>Not required</td></tr>
+<tr><td>enabled-audio-track-ids</td><td>-</td>                                          <td>Required</td></tr>
+<tr><td>select-video-track-id</td>  <td>-</td>                                          <td>Required</td></tr>
```
pthatcherg: This should be selected-video-track-id, as it is in the .cddl file.  Could you change it (and the other place in this file, which is a mistake)?

index.bs
```diff
@@ -1234,7 +1231,25 @@ intentionally mirror settable attributes and methods of the
...
+<tr><td>added-text-tracks</td>      <td>-</td>                                          <td>Not required</td></tr>
+<tr><td>changed-text-tracks</td>    <td>-</td>                                          <td>Not required</td></tr>
```
pthatcherg: Perhaps "none" or "empty" instead of "-" would be more clear.

index.bs
```diff
@@ -1234,7 +1231,25 @@ intentionally mirror settable attributes and methods of the
...
+<tr><td>preload</td>                <td>Decided by the receiver</td>                    <td>Not required</td></tr>
+<tr><td>loop</td>                   <td>False</td>                                      <td>Required</td></tr>
+<tr><td>paused</td>                 <td>Decided by the receiver</td>                    <td>Required</td></tr>
+<tr><td>muted</td>                  <td>Decided by the receiver</td>                    <td>Required</td></tr>
+<tr><td>volume</td>                 <td>Decided by the receiver</td>                    <td>Required</td></tr>
+<tr><td>seek</td>                   <td>-</td>                                          <td>Required</td></tr>
+<tr><td>fast-seek</td>              <td>-</td>                                          <td>Required</td></tr>
+<tr><td>playback-rate</td>          <td>Decided by the receiver</td>                    <td>Not required</td></tr>
+<tr><td>poster</td>                 <td>Decided by the receiver</td>                    <td>Not required</td></tr>
+<tr><td>enabled-audio-track-ids</td><td>-</td>                                          <td>Required</td></tr>
```
pthatcherg: These IDs are for the *controls*, not the *state*.  There are 2 different messages.   Maybe we should have a table for both, but the issue/TODO was for the state table, so we need at least that.

index.bs
```diff
@@ -1234,7 +1231,25 @@ intentionally mirror settable attributes and methods of the
     if a cue ID is invalid (removing an un-added ID or adding an ID twice, for example),
     the receiver may reject the text track change.
 
-Issue: Add a table for whether it's required and what the default is.
+<div class="note">
+<table>
+<tr><th>Field</th>                  <th>Default value</th>                              <th>Receiver support</th></tr>
```
pthatcherg: Should make it clear this is the default for the *initial* controls.  For all other controls, the default is "unchanged".

takumif: Updated.

pthatcherg:
I think this is good enough to merge.  We can tweak it further if necessary later.

PR #224, opened 2019-09-16
Add support for extended mime types and CSS media queries

pthatcherg:
Resolves 2 issues from Berlin f2f:
- Add extended mime types to remote playback 
- Add CSS media query to remote playback

***
<a href="https://pr-preview.s3.amazonaws.com/webscreens/openscreenprotocol/pull/224.html" title="Last updated on Sep 19, 2019, 7:32 AM UTC (b2d0c15)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webscreens/openscreenprotocol/224/d4526c9...b2d0c15.html" title="Last updated on Sep 19, 2019, 7:32 AM UTC (b2d0c15)">Diff</a>


index.bs
```diff
@@ -1104,9 +1105,13 @@ values:
 Note: A version 4 (pseudorandom) [=UUID=] is recommended as it meets the
 requirements for a remote-playback-id.
 
-: urls
-:: The [=media resources=] that the controller has selected for playback on the
-    receiver.
+: sources The [=media resources=] that the controller has selected for playback on the
+    receiver.  Each source must include a URL
+    (see [HTMLSourceElement.src](https://html.spec.whatwg.org/multipage/embedded-content.html#dom-source-src))
```
mfoltzgoogle: Can you use Bikeshed autolinking, like {{HtmlMediaElement#src}}?

https://tabatkins.github.io/bikeshed/#autolink-microsyntax

messages_appendix.cddl
```diff
@@ -266,13 +266,19 @@ remote-playback-availability-event = {
...
+  2: [* remote-playback-source] ; sources
   ? 3: [* text] ; text-track-urls
   ? 4: [* http-header] ; headers
   ? 5: remote-playback-controls ; controls
   ? 6: {streaming-session-start-request-params} ; remoting
 }
 
+remote-playback-source = {
+  1: text; url
+  2: text; css-media-query
```
mfoltzgoogle: .media and .type are both optional for HtmlMediaElement, right?
So the message fields should also be optional I think.

mfoltzgoogle: My reading of HTML is that the media query is only applicable if the `<source>` element is being used to select an image file in a `<picture>`.  So this can be dropped.

https://html.spec.whatwg.org/multipage/embedded-content.html#the-source-element

mfoltzgoogle: I think it's okay to make extended-mime-type mandatory; per the WebIDL, an omitted attribute gets reflected to script as an empty string.

messages_appendix.cddl
```diff
@@ -332,7 +338,7 @@ remote-playback-state-event = {
 remote-playback-id = uint
 
 remote-playback-controls = {
-  ? 1: url ; source
+  ? 1: text ; source-url
```
mfoltzgoogle: Are these changes related?  I think I prefer the "url" semantic type for these fields.

messages_appendix.cddl
mfoltzgoogle: I updated the PR to reflect this.

mfoltzgoogle:
I fixed the linking and dropped the media queries.  PTAL.

messages_appendix.cddl
pthatcherg: "url" is a type that we used to have be removed at some point.  Whenever we removed it, we made a mistake here in not changing this.

If we want to re-add the url type, we can do that in a separate RP.

PR #225, opened 2019-09-16
Reference Media Capabilities ColorGamut for color space capability

pthatcherg:
Reference CSS Color Module Level 4 for color spaces

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/openscreenprotocol/pull/225.html" title="Last updated on Oct 27, 2021, 10:49 PM UTC (2585352)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/openscreenprotocol/225/7305111...2585352.html" title="Last updated on Oct 27, 2021, 10:49 PM UTC (2585352)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/openscreenprotocol/pull/225.html" title="Last updated on Jan 21, 2022, 1:19 AM UTC (ceb11ae)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/openscreenprotocol/225/3beedbd...ceb11ae.html" title="Last updated on Jan 21, 2022, 1:19 AM UTC (ceb11ae)">Diff</a>


mwatson2:
It might be better to look to [media capabilities](https://w3c.github.io/media-capabilities/) rather than CSS if you are interested in expressing video capabilities. CSS does not yet support HDR, for example.

index.bs
```diff
@@ -1702,10 +1702,12 @@ following additional fields:
     display could return this value as 1.6 to indicate its preferred content
     scaling. Default is none.
 
-: color-profiles (optional)
-:: An optional field indicating what color profiles are understood.
-    The listener may use these values to determine how to encode
-    video. Some examples include: sRGBv4, Rec709, DciP3. The default value is sRGBv4.
+: color-spaces (optional)
+:: An optional field indicating what color spaces are understood.
+    The listener may use these values to determine how to encode video. Valid
```
mfoltzgoogle: Do you mean "media receiver" instead of listener?

index.bs
```diff
@@ -1702,10 +1702,12 @@ following additional fields:
...
+    values are defined by [CSS Color Module Level
+    4](https://www.w3.org/TR/css-color-4/#predefined). The default value is
```
mfoltzgoogle: Can you use Bikeshed autolinks here?  That will add CSS4 as a reference.

https://tabatkins.github.io/bikeshed/#autolink-biblio

index.bs
mfoltzgoogle: Addressed in the updated PR.

index.bs
mfoltzgoogle: Now using ColorGamut from Media Capabilities API.

mfoltzgoogle:
This PR has been updated to refer instead to the [ColorGamut enum](https://www.w3.org/TR/media-capabilities/#colorgamut) from MediaCapabilities.  As is discussed at length in w3c/media-capabilities#14, w3c/media-capabilities#118, w3c/media-capabilities#119 and the w3c/media-capabilities#124 PR, this is intended to model the color spaces that are understood by a video decoder.

Along with HDR metadata and transfer function, this should give the media sender sufficient information to send an HDR video track and have it understood by the media receiver.

With this change, requesting the capabilities of the actual display on the video receiver for various color profiles/color spaces would have to be done as a separate protocol feature.

mfoltzgoogle (approved):

index.bs
```diff
@@ -1683,10 +1683,12 @@ following additional fields:
...
-: color-profiles (optional)
-:: An optional field indicating what color profiles are understood.
-    The listener may use these values to determine how to encode
-    video. Some examples include: sRGBv4, Rec709, DciP3. The default value is sRGBv4.
+: color-gamut (optional)
+:: An optional field indicating what color spaces can be understood by the media
+    receiver. The media sender may use these values to determine how to encode
+    video. Valid values correspond to
+    [[MEDIA-CAPABILITIES#colorgamut|ColorGamut]] in the Media Capabilities API.
+    The default value is "srgb".
```
tidoust, line 1691:
```suggestion
    The default value is "srgb". [[!MEDIA-CAPABILITIES]]
```

Use of Bikeshed's cross-spec section link is not enough to add Media Capabilities to the list of normative references. There needs to be an explicit bibliographical reference.

tidoust (approved):

mfoltzgoogle:
Tagging @chcunningham who offered to check this PR before landing :)

chcunningham:
Values look correct. 

FYI, MediaCapabilities chose these color gamut properties to align with CSS color-gamut https://drafts.csswg.org/mediaqueries/#color-gamut. But I'm totally flexible on whatever spec you want to reference. 

Does "understood by the media receiver" mean the video pipeline understands the color space, or does it also mean display is capable of outputting in that color space? For MediaCapabilties, we are describing only the former. Callers are expected to additionally check CSS color-gamut to determine the display's capabilities. I'm not sure what makes the most sense for this API.

For MediaCapabilities, colorGamut is just one of 3 properties needed to support for a given flavor of HDR. The others are [TransferFunction](https://www.w3.org/TR/media-capabilities/#enumdef-transferfunction) and [HdrMetadataType](https://www.w3.org/TR/media-capabilities/#hdrmetadatatype). Do those make sense as additional parameters in this spec?

mfoltzgoogle:
Thanks for the comments @chcunningham!

> Does "understood by the media receiver" mean the video pipeline understands the color space, or does it also mean display is capable of outputting in that color space? 

We were largely waiting for resolution in the Media WG for how to model the difference between the two.  It appears that w3c/media-capabilities#119 was closed by the addition of `color-gamut` to CSS Media Queries 5 (using the same values as Media Capabilities).

The question now, is does the sender need to know both the decoding and rendering color-gamut before it decides to send an HDR stream to the receiver?  I would argue "no", because if the receiver's display was not capable of rendering a specific color-gamut, then there would be no reason to advertise its decoding capability for a given streaming session.

Advertisement via capabilities should mean that the receiver is both capable of receiving that color-gamut, and configured to display.  I'll update the PR to clarify.

> For MediaCapabilities, colorGamut is just one of 3 properties needed to support for a given flavor of HDR. The others are TransferFunction and HdrMetadataType. Do those make sense as additional parameters in this spec?

Yes they would, if they affect how the sender might want to encode media (or decide if a media stream was compatible with a receiver).  I'd like to tackle that in a separate PR since conveying the HDR data over the wire could require additional changes to the protocol.

chcunningham:
> I would argue "no", because if the receiver's display was not capable of rendering a specific color-gamut, then there would be no reason to advertise its decoding capability for a given streaming session.

SGTM. One thing to point out: for MC we chose to avoid display capabilities in part because you can have multiple displays with distinct capabilities. CSS handles this gracefully. But I could see this not being an issue for your use case - do you assume receivers have just one screen?

> I'd like to tackle that in a separate PR since conveying the HDR data over the wire could require additional changes to the protocol.

SGTM

mfoltzgoogle:
> But I could see this not being an issue for your use case - do you assume receivers have just one screen?

The protocol doesn't have a way to target a stream to a specific screen, so I would say that is the case.  If there are multiple screens with different capabilities, they could be represented as different agents (endpoints).  For the majority of smart TV/presentation device scenarios, this should be fine.

mfoltzgoogle:
Leaving this open for a few more days in case folks want to look at the current version of the PR.

Issue #241, opened 2020-01-31
Add a back pressure signal for media remoting


takumif:
Per the [TPAC discussion](https://www.w3.org/2019/09/16-webscreens-minutes.html#x02), when media remoting for Remote Playback, there should be a back-pressure signal to indicate whether or not the media bitrate is too high given the network/receiver constraints (e.g. bandwidth, buffer size).

w3c/remote-playback#132 tracks this on the Remote Playback API side.

mfoltzgoogle:
From https://www.w3.org/2019/09/16-webscreens-minutes.html#x02:

> RESOLVED: Add a backpressure signal to the OSP

Resolving this, should close w3c/remote-playback#132.

anssiko:
>ACTION: People to review the PR https://github.com/w3c/openscreenprotocol/pull/248 and provide feedback

via https://www.w3.org/2020/10/20-webscreens-minutes.html#a02

mfoltzgoogle:
See the conversation thread on PR #248 for more background/discussion of this issue.

PR #248, opened 2020-10-19
Adds a backpressure signal to the streaming protocol.

mfoltzgoogle:
Addresses Issue #241: Add a back pressure signal for media remoting.

This adds a backpressure signal to the `streaming-session-receiver-stats-event` message, so the receiver can report the status of its frame buffer (if it is using one).  The states are borrowed from my interpretation of https://github.com/w3c/remote-playback/pull/126.

An item to discuss is how and whether to expose this to script along the lines of  https://github.com/w3c/remote-playback/pull/126.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/openscreenprotocol/pull/248.html" title="Last updated on Mar 16, 2021, 9:33 PM UTC (0673f33)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/openscreenprotocol/248/c297653...0673f33.html" title="Last updated on Mar 16, 2021, 9:33 PM UTC (0673f33)">Diff</a>


mfoltzgoogle:
@takumif @mounirlamouri @eric-carlson @chrisn 

Notifying folks who were involved in earlier discussions around this for any feedback on the PR.

The idea is to allow the receiver of a media stream (i.e., from media remoting) to signal the sender of the media stream whether it should adjust its data rate up or down to prevent buffer overruns or underruns.  Here's the timeline of previous discussions:

- Initial proposal for the Remote Playback API by @takumif: w3c/remote-playback#126
- Discussion at TPAC 2019: https://www.w3.org/2019/09/16-webscreens-minutes.html#x02
- Resulting OSP issue to add a backpressure signal:  #241
- Discussion at TPAC 2020: https://www.w3.org/2020/10/20-webscreens-minutes.html#a02
- Followup issue to expose this signal via the Remote Playback API (in the future): w3c/remote-playback#132

louaybassbouss:
Thanks @mfoltzgoogle for the PR happy to review it but I need to understand one thing: the main use case as I can see from the related issues is MSE. Are the frames decoded on the sender before they are transmitted to the receiver or the sender just transmits the media segments to the receiver without re-encoding?

louaybassbouss:
@mfoltzgoogle the backpressure solution you provided LGTM. It added @chrisn and @mounirlamouri as reviewers and I hope we can merge it soon. @takumif @eric-carlson feel free to review as well (I couldn't add you as reviewer)

mfoltzgoogle:
> Thanks @mfoltzgoogle for the PR happy to review it but I need to understand one thing: the main use case as I can see from the related issues is MSE. Are the frames decoded on the sender before they are transmitted to the receiver or the sender just transmits the media segments to the receiver without re-encoding?

That's correct, the main use case is when the sender has pre-encoded media that is being fetched from a server or file, and wants to play it back on a remote device (agent receiving the stream in the protocol).  For Web apps using the Remote Playback API, the media would be provided from the Web app to the protocol through MSE.

The sender can use the backpressure signal from the receiver to make adjustments to the send rate and how much it's buffering on the sender side.  Some senders may be able to transcode the media to change the effective data rate.

It might be helpful to generate simulations showing when the signal might be sent and how it can be used to ensure uninterrupted playback on the receiver device when remoting from MSE.

index.bs
```diff
@@ -2121,7 +2121,26 @@ Video encoding sender stats include the following fields:
 During a streaming session, the receiver should send stats with the
 [=streaming-session-receiver-stats-event=] at the interval the sender requested.
 It should send all of the following stats for all of the media streams it is
-receiving.  The [=streaming-session-receiver-stats-event=] message contains the
+receiving.
+
+If the receiver is using a buffer to hold frames before playing them out, it
+should also sends the status of that buffer using the `remote-buffer-status` field.
```
chrisn, line 2127: typo: sends -> send

chrisn (approved):
Apart from one minor typo, this looks good to me. I agreed with @mfoltzgoogle's comment in the meeting that feedback from prototyping could be helpful.

louaybassbouss:
Thanks @mfoltzgoogle for the explanation and @chrisn for review. Lets wait 1-2 days for more feedback and merge this PR.

anssiko:
LGTM.

We did not explicitly note #241 as a FPWD blocker in https://github.com/w3c/openscreenprotocol/issues/249#issuecomment-784313022 but seems like we want to squeeze this in the FPWD so I’d suggest snapshotting that release when this PR is merged.

@tidoust, can you check we can make 18 March FPWD with this integrated i.e. builds w/o errors and passes all the publication checks? Thanks!

index.bs
mfoltzgoogle: Done.

mfoltzgoogle:
It sounds like we want to ensure this is in the FPWD, so I went ahead and merged the current snapshot of the PR.  I will leave the issue open in case there is additional feedback from reviewers.

Issue #259, opened 2021-03-01
travis-ci jobs to commit index.html changes are failing


mfoltzgoogle:
It looks like since October, the travis-ci jobs to update our generated spec (on the gh-pages branch) when pull requests are merged is failing.

Example today: https://travis-ci.com/github/w3c/openscreenprotocol/builds/218631036

What is failing is the step in [deploy.sh](https://github.com/w3c/openscreenprotocol/blob/master/scripts/deploy.sh) to extract the encryption keys from Travis environment variables.

```bash
+ENCRYPTED_KEY_VAR=encrypted_6c9ad971e3e1_key
+ENCRYPTED_IV_VAR=encrypted_6c9ad971e3e1_iv
+ENCRYPTED_KEY=
+ENCRYPTED_IV=
+openssl aes-256-cbc -K -iv -in ../scripts/deploy_key.enc -out ../scripts/deploy_key -d
iv undefined
```

It's been too long since I set this up to remember how to fix it.  Maybe @tidoust is familiar with how other repos are doing this and can point to a working script?

mfoltzgoogle:
It looks like Travis isn't setting the environment variables with the encryption key/IV for the GitHub key that the script needs to push commits to the repo.

@tidoust, it looks like you set this up originally in PR #230.  Do you mind seeing if you can re-generate these keys with Travis, or send me instructions on how to do so?

tidoust:
TLDR: I enabled a temporary workaround but I need to re-generate the keys and `deploy_key.enc` for a proper fix.

A longer explanation:
That is because there are two travis-ci domains in practice for the same service, which is something I realized only recently. There is the old `travis-ci.org` which was triggered through a Webhook, and the new `travis.com`, which is enabled at the W3C organization level on GitHub, and does not require any Webhook.

When the repo transitioned from WICG to W3C, the environment variables attached to the `travis-ci.org` repo configuration should have been automatically migrated to `travis-ci.com` (at least according to the [documentation](https://docs.travis-ci.com/user/migrate/open-source-repository-migration/#what-information-will-be-transferred-to-travis-cicom)). For some reason, that did not happen. I may have pushed the wrong button at the time (I don't recall having pushed any Travis-CI button...).

The Webhook remained in place for some time, and two Travis CI jobs were run in parallel each time a push was made: one handled by `travis-ci.org`, which succeeded, and one handled by `travis-ci.com`, which failed because the environment variables had not been copied over.

[...]

PR #262, opened 2021-03-02
Switch build to GitHub action using spec-prod

tidoust:
This creates a GitHub action that replaces the Travis CI worfklow for re-generating the specification and updating the `gh-pages` branch.

The action uses w3c/spec-prod under the hoods:
https://github.com/w3c/spec-prod/#spec-prod

w3c/spec-prod takes care of everything and also validates the spec (markup and broken hyperlinks).

Good thing about this approach is that no specific secret tokens are needed, since GitHub actions are allowed to push to the gh-pages branch. Also, no need to maintain a complex deploy script.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/openscreenprotocol/pull/262.html" title="Last updated on Mar 3, 2021, 9:01 AM UTC (3d71847)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/openscreenprotocol/262/334c104...3d71847.html" title="Last updated on Mar 3, 2021, 9:01 AM UTC (3d71847)">Diff</a>


tidoust:
So, build currently fails because HTML validation fails. Markup is invalid due to the CDDL messages appendix that has the structure `<pre>... <p> </p> </pre>`, and `<p>` is not allowed in `<pre>`. I do not know yet where these `<p>` are created, since they do not exist in `messages_appendix.html`. I'll investigate (this needs to be fixed for publication as First Public Working Draft) and will update the PR accordingly.

tidoust:
OK, I found the right include setting in Bikeshed to avoid the introduction of `<p>` tags.

All things should work now and this PR should fix #259.

.github/workflows/pr-push.yml
```diff
@@ -0,0 +1,31 @@
...
+  push:
+    branches: [master]
+  workflow_dispatch:
+
+jobs:
+  main:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
```
mfoltzgoogle, line 13: I am not familiar with GitHub's workflow syntax.  Where are these actions defined?  Can we include a link to their definitions, either as a comment here or in a README.md file?

.github/workflows/pr-push.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Install Python dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install Pygments
+
+      - name: Generate messages_appendix.html
+        run: python scripts/pygmentize_dir.py
+
+      - name: Build and validate index.html, push to gh-pages branch if needed
+        uses: w3c/spec-prod@v2
```
mfoltzgoogle, line 29: I'm assuming this is a reference to a workflow defined in the w3c/spec-prod repository?  Can a link to it (or its documentation) be included?

index.bs
```diff
@@ -195,7 +195,8 @@ Non-Functional Requirements {#requirements-non-functional}
 1.  It should be possible to implement an OSP agent using modest
     hardware requirements, similar to what is found in a low end smartphone,
     smart TV or streaming device. See the [Device
-    Specifications](device_specs.md) document for agent hardware specifications.
+    Specifications](https://github.com/w3c/openscreenprotocol/blob/master/device_specs.md)
```
mfoltzgoogle, line 198: Maybe these links should point to the gh-pages branch since we are serving the spec document from that branch?

I don't feel strongly since the two branches should always be kept in sync by the workflow.

mfoltzgoogle (approved):
Thanks for making these changes; it would be helpful for me to document the referenced actions/workflows for future maintenance :-)

.github/workflows/pr-push.yml
mfoltzgoogle: A link to https://github.com/w3c/spec-prod/#spec-prod in a comment would be perfect.

.github/workflows/pr-push.yml
mfoltzgoogle: A link to https://github.com/actions/checkout/blob/main/README.md would work here.

index.bs
tidoust: I would keep the `gh-pages` name as a "hidden" one: no need to expose the name of the branch that gets published as GitHub Pages, to avoid breaking links if we change the name. Actually, that applies to `master` as well, which we'll soon need to rename to `main` (we're progressively transitioning all W3C repositories to use `main` as the default branch instead of `master`).

I thought something had to be done for GitHub to publish Markdown files as HTML but I was wrong. It's all being done by default, you just have to replace the `.md` extension with `.html`. I pushed an update that targets the HTML pages under `w3c.github.io` (i.e. [capabilities.html](https://w3c.github.io/openscreenprotocol/capabilities.html) and [device_specs.html](https://w3c.github.io/openscreenprotocol/device_specs.html)). We can always choose a different theme and adjust the header and footer on these pages later on if we feel that's needed.

mfoltzgoogle:
Looks good, and thanks again @tidoust.  Feel free to merge this when you are ready.

PR #168, opened 2019-05-22
Draft of Open Screen Protocol explainer document

mfoltzgoogle:
Addresses Issue #165: Draft TAG Explainer.

This is a draft of an explainer for the Open Screen Protocol, largely following the outline here:

https://github.com/w3ctag/w3ctag.github.io/blob/master/explainers.md

There are a couple of things different about this one.  Because this is not a Web API, the "sample code" is not intended to be developer facing, but instead is a session showing what kinds of data are exchanged.  I didn't finish this section yet, but can later based on feedback from the group.

I would also like to add more detailed use cases and problems (incompatibilities) that should be addressed by OSP in a future PR.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/openscreenprotocol/pull/168.html" title="Last updated on Mar 12, 2021, 10:48 PM UTC (f671bb1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/openscreenprotocol/168/c1a8b35...f671bb1.html" title="Last updated on Mar 12, 2021, 10:48 PM UTC (f671bb1)">Diff</a>

PR #270, opened 2021-03-12
Update README.md

mfoltzgoogle:
This updates the README.md file to reflect the current status of the spec.   It previously mentioned that we were still considering WebRTC, etc. for the protocol.  Instead it lists the chosen technologies, adds a bit about the current status, and links to the explainer in PR #168.


anssiko (approved):
This looks great, thanks!

Issue #273, opened 2021-03-25
[mDNS] OpenScreenProtocol should not restrict DNS-SD to mDNS


DavidSchinazi:
The spec currently states:
> Open Screen Protocol agents must discover one another using DNS-SD over mDNS. To do so, OSP agents must use the Service Name _openscreen._udp.local.

This limitation seems arbitrary and will prevent OSP from working in enterprise contexts. The non-mDNS-based DNS-SD design allows service discovery to work even across broadcast domains. In particular, it's common for enterprises to have their user-facing Wi-Fi on a separate broadcast domain from their ethernet-connected hardware (printers, TVs, etc.). That prevents mDNS from working but a system administrator can decide to setup DNS-SD to allow devices on the Wi-Fi to discover specific services on other parts of the network - this is mainly done for printers today, but OSP shouldn't artificially prevent itself from benefiting from this technology.

louaybassbouss:
Thanks @DavidSchinazi for your feedback can you provide more background information of the "non-mDNS" approach you have in mind? 
PS: I recommend to check the [archive/mdns.md](https://github.com/w3c/openscreenprotocol/blob/master/archive/mdns.md) document which includes details about our investigation (@mfoltzgoogle  FYI)

DavidSchinazi:
Hi @louaybassbouss! Thanks for the link to your analysis, I think it's very thorough and complete apart from the fact that it conflates mDNS and DNS-SD. mDNS (which is defined in [RFC 6762](https://datatracker.ietf.org/doc/html/rfc6762)) is a subset of DNS-SD (which is defined in [RFC 6763](https://datatracker.ietf.org/doc/html/rfc6763)). DNS-SD works over both mDNS and non-mDNS (i.e. regular unicast DNS). Going into a bit more detail, DNS-SD has a concept of domain which means where you are looking for a service. The domain `local.` is reserved as a [special-use domain name](https://datatracker.ietf.org/doc/html/rfc6761): DNS-SD queries in the `local.` domain are sent over mDNS whereas queries for all other domains are set over regular unicast DNS. All of the properties that the Open Screen Protocol cares about are provided by DNS-SD, with mDNS being a subset.

mfoltzgoogle:
[...]

PR #283, opened 2021-04-13
Allow additional discovery mechanisms.

mfoltzgoogle:
Addresses Issue #273: [mDNS] OpenScreenProtocol should not restrict DNS-SD to mDNS

This makes it explicit that while DNS-SD over mDNS is mandatory to implement, the rest of the protocol doesn't rely on mDNS being the only way for agents to discover one another, and they can query DNS-SD over unicast DNS.

Note that we may need to adjust the `at` mechanism intended to restrict connection attempts to intra-LAN scenarios.  Would like to look at that in a separate issue.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -254,9 +254,14 @@ Non-Functional Requirements {#requirements-non-functional}
 Discovery with mDNS {#discovery}
 ===============================
 
-[=Open Screen Protocol agents=] must discover one another using
-[[RFC6763|DNS-SD]] over [[RFC6762|mDNS]].  To do so, OSP agents must use the
-[=Service Name=] `_openscreen._udp.local`.
+[=Open Screen Protocol agents=] discover one another by advertising and
+listening for information identifying themselves along with an IP service
+endpoint.  Agent advertisment and discovery through [[RFC6763|DNS-SD]] and
+[[RFC6762|mDNS]] is defined by this specification and is mandatory to implement
+by all agents. However, agents are free to implement additional discovery
+mechanisms, such as querying for the same DNS-SD records via unicast DNS.
+
+OSP agents must use the DNS-SD [=Service Name=] `_openscreen._udp.local`.
 
 Issue(107): Define suspend and resume behavior for discovery protocol.
 


mfoltzgoogle:
@DavidSchinazi here's some updated language about DNS-SD usage for OSP.

index.bs
```diff
@@ -254,9 +254,14 @@ Non-Functional Requirements {#requirements-non-functional}
 Discovery with mDNS {#discovery}
 ===============================
 
-[=Open Screen Protocol agents=] must discover one another using
-[[RFC6763|DNS-SD]] over [[RFC6762|mDNS]].  To do so, OSP agents must use the
-[=Service Name=] `_openscreen._udp.local`.
+[=Open Screen Protocol agents=] discover one another by advertising and
+listening for information identifying themselves along with an IP service
+endpoint.  Agent advertisment and discovery through [[RFC6763|DNS-SD]] and
```
DavidSchinazi, line 259:
```suggestion
endpoint.  Agent advertisement and discovery through [[RFC6763|DNS-SD]] and
```

index.bs
```diff
@@ -254,9 +254,14 @@ Non-Functional Requirements {#requirements-non-functional}
...
+[[RFC6762|mDNS]] is defined by this specification and is mandatory to implement
+by all agents. However, agents are free to implement additional discovery
+mechanisms, such as querying for the same DNS-SD records via unicast DNS.
+
+OSP agents must use the DNS-SD [=Service Name=] `_openscreen._udp.local`.
```
DavidSchinazi, line 264: The Service Name is `_openscreen._udp` for both DNSSD and mDNS. The next component is the Domain, and that's `local` in mDNS and can be anything else in DNSSD. I'd suggest changing this instance and others in this file. It's good for `local` to show up in examples but I'd avoid it in normative text because that effectively disallows unicast DNS.

```suggestion
OSP agents must use the DNS-SD [=Service Name=] `_openscreen._udp`.
```

DavidSchinazi (commented):
Thanks for doing this!

index.bs
louaybassbouss: Since mDNS is mandatory and unicast DNS is additional/optional maybe we can make it clearer.   

_OSP agents must use the DNS-SD [=Service Name=] `_openscreen._udp.local` for mDNS and `_openscreen._udp.*` if unicast DNS is supported._ 

Maybe the `*` in  `_openscreen._udp.*` needs to be replaced with something concrete? wdyt?

DavidSchinazi: @louaybassbouss sorry to be slightly pedantic here, but `_openscreen._udp.local` is not a service name. In that example, `_openscreen._udp` is the service name and `local` is the domain. The term "service name" is defined in [RFC 6763 Section 4.1](https://datatracker.ietf.org/doc/html/rfc6763#section-4.1).

louaybassbouss: @DavidSchinazi thanks for clarification LGTM.

louaybassbouss (approved):
Thanks @mfoltzgoogle  for this PR LGTM also changes proposed by @DavidSchinazi

Issue #280, opened 2021-04-06
Clarify the supported signature algorithms for certificates


sleevi:
Separate from #277, the support for signature algorithms currently states: https://github.com/w3c/openscreenprotocol/blob/5488c7b7cce9c9c64ba97348a8deb0c6c50eb9bb/index.bs#L367-L370

However, this is not a defined X.509/PKIX signature algorithm. It's likely these names were derived from [RFC 8446, Section 4.2.3](https://tools.ietf.org/html/rfc8446#section-4.2.3)'s `SignatureScheme` registry, but those don't directly translate here.

Further, experience has shown that implementors who have attempted to implement ECDSA within certificates have, at times, incorrectly encoded the `subjectPublicKeyInfo`/`signatureAlgorithm`, through incorrect interpretation of the algorithm parameters, or through novel (problematic) attempts at representing the curve.

In more modern certificate profiles, an effort has been made to explicitly specify the encoded byte representation, namely:
* `secp256r1_sha256` = `301306072a8648ce3d020106082a8648ce3d030107` for the public key `AlgorithmIdentifier`, `300a06082a8648ce3d040302` for the signature `AlgorithmIdentifier`
* `secp384r1_sha384` = `301006072a8648ce3d020106052b81040022` for the public key `AlgorithmIdentifier`, `300a06082a8648ce3d040303` for the signature `AlgorithmIdentifier`

mfoltzgoogle:
Closed by #288.

Issue #279, opened 2021-04-06
The keyUsage name is digitalSignature, not signing


sleevi:
The current certificate profile states:
https://github.com/w3c/openscreenprotocol/blob/5488c7b7cce9c9c64ba97348a8deb0c6c50eb9bb/index.bs#L412-L415

The `Signing` key usage should refer to the precise key usage defined in [RFC 5280, Section 4.2.1.3](https://tools.ietf.org/html/rfc5280#section-4.2.1.3), which is `digitalSignature`

mfoltzgoogle:
Closed by #288.

Issue #277, opened 2021-04-06
Consider removing support for P-521


sleevi:
A number of user agents intentionally decline to implement P-521, due to the significant performance overheads incurred relative to the limited security benefit provided in practice. For example, both Chrome and Firefox prohibit such certificates within TLS connections.

The current language includes P-521 as optional, as captured at https://github.com/w3c/openscreenprotocol/blob/5488c7b7cce9c9c64ba97348a8deb0c6c50eb9bb/index.bs#L363-L371

This will likely exacerbate interoperability issues, and also arguably incompatible with the goal stated in Non-Functional Requirements, namely https://github.com/w3c/openscreenprotocol/blob/5488c7b7cce9c9c64ba97348a8deb0c6c50eb9bb/index.bs#L206-L210

Suggestion: Remove P-521.

mfoltzgoogle:
Closed by #295.

PR #288, opened 2021-10-27
Fixes algorithm and signing fields in agent certificate.

mfoltzgoogle:
Addresses:
  *  Issue #279: The keyUsage name is digitalSignature, not signing
  *  Issue #280: Clarify the supported signature algorithms for certificates

For the first item, it explicitly references the `keyUsage` bit for `digitalSigning`.

For the second item, I added a table that matches the TLS signature_scheme values to the `AlgorithmIdentifier` values that should be used in the certificate.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/openscreenprotocol/pull/288.html" title="Last updated on Oct 27, 2021, 6:25 AM UTC (f96a9d5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/openscreenprotocol/288/7305111...f96a9d5.html" title="Last updated on Oct 27, 2021, 6:25 AM UTC (f96a9d5)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/openscreenprotocol/pull/288.html" title="Last updated on Nov 10, 2022, 1:31 AM UTC (d62bf1b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/openscreenprotocol/288/246bc07...d62bf1b.html" title="Last updated on Nov 10, 2022, 1:31 AM UTC (d62bf1b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -31,8 +31,6 @@ urlPrefix: https://www.w3.org/TR/presentation-api/#dfn-; type: dfn; spec: PRESEN
     text: presentation request url
     text: receiving user agent
     text: creating a receiving browsing context
-urlPrefix: https://www.w3.org/TR/presentation-api/; type: interface; spec: PRESENTATION-API
-    text: PresentationConnection
 urlPrefix: https://w3c.github.io/remote-playback/#dfn-; type: dfn; spec: REMOTE-PLAYBACK
     text: availability sources set
     text: compatible remote playback device
@@ -1449,13 +1447,13 @@ changed.
 : loading
 :: The state of network activity for loading the [=media resource=]. See
     {{HTMLMediaElement/networkState|HTMLMediaElement.networkState}}.
-    The default is empty ({{NETWORK_EMPTY}}
+    The default is empty ({{NETWORK_EMPTY}})
     for the initial state in the [=remote-playback-start-response=] message.
 
 : loaded
 :: The state of the loaded media (whether enough is loaded to play). See
     {{HTMLMediaElement/readyState|HTMLMediaElement.readyState}}.
-    The default is nothing ({{HAVE_NOTHING}}
+    The default is nothing ({{HAVE_NOTHING}})
     for the initial state in the [=remote-playback-start-response=] message.
 
 : error
@@ -1631,10 +1629,10 @@ This is how the [[REMOTE-PLAYBACK|Remote Playback API]] uses the
 messages defined in [[#remote-playback-protocol]]:
 
 When [[REMOTE-PLAYBACK#the-list-of-available-remote-playback-devices|section
-6.2.1.2]] says "This list contains [=remote playback devices=] and is populated
+5.2.1.2]] says "This list contains [=remote playback devices=] and is populated
 based on an implementation specific discovery mechanism" and
 [[REMOTE-PLAYBACK#the-list-of-available-remote-playback-devices|section
-6.2.1.4]] says "Retrieve available remote playback devices (using an
+5.2.1.4]] says "Retrieve available remote playback devices (using an
 implementation specific mechanism)", the user agent may use the mDNS, QUIC,
 [=agent-info-request=], and [=remote-playback-availability-request=] messages
 defined previously in this spec to discover [=receivers=].  The
@@ -1643,7 +1641,7 @@ sources set=].
 
 When
 [[REMOTE-PLAYBACK#establishing-a-connection-with-a-remote-playback-device|section
-6.2.4]] says "Request connection of remote to device. The implementation of this
+5.2.4]] says "Request connection of remote to device. The implementation of this
 step is specific to the user agent." and "Synchronize the current media element
 state with the remote playback state", the controller may send the
 [=remote-playback-start-request=] message to the receiver to start remote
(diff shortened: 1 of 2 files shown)


mfoltzgoogle:
I believe this PR is still ready for review.  We can review it next week as it's been open a while.

index.bs
```diff
@@ -367,16 +369,43 @@ QUIC connection.
...
+</tr>
+<tr>
+  <td>`secp384r1_sha384`</td>
+  <td>`301006072a8648ce3d020106052b81040022`</td>
+  <td>`300a06082a8648ce3d040303`</td>
+</tr>
+<tr>
+  <td>`secp521r1_sha512`</td>
+  <td>TBD</td>
+  <td>TBD</td>
```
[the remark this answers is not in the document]
chrisn: I also found some discussion in https://groups.google.com/g/mozilla.dev.security.policy/c/7O34-DmZeC8, so I guess there's a question about whether to include this at all? (I'm no expert here, so would appreciate Ryan's input)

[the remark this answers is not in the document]
mfoltzgoogle: Thanks Chris for digging up that thread; based on both that thread, and comments by @sleevi on #277, I've removed it from this PR.  I'll put up another PR to remove the remaining refrences to P521.

chrisn (approved):

index.bs
```diff
@@ -367,16 +369,38 @@ QUIC connection.
...
-* 256-bit, 384-bit, or 521-bit ECDSA public key
-* Self-signed
-* Supporting the at least one of the following signature algorithms:
-    * secp256r1_sha256
-    * secp384r1_sha384
-    * secp521r1_sha512
-* Valid for signing
+* 256-bit, 384-bit, or 521-bit ECDSA public key.
+* Self-signed.
+* Supporting the at least one of the [=certificate algorithms=] listed below.
```
tidoust, line 374: "The at least" seems to be used in legal documents, but I don't think that's a common construct in technical specs.

```suggestion
* Supporting at least one of the [=certificate algorithms=] listed below.
```

index.bs
```diff
@@ -367,16 +369,38 @@ QUIC connection.
...
+    * The `AlgorithmIdentifiers` are defined in [[!RFC5480]].
```
tidoust, line 375: I confess I'm lost when I read this and the table below. RFC5480 does not seem to define the public key and signature numbers that appear in the table.

If it explains how to compute these numbers, I would clarify with:

```suggestion
    * Public key and signature `AlgorithmIdentifiers` are derived from [[!RFC5480]].
```

If the numbers that appear in the table come from elsewhere, it would be good to reference the source.

Or did we just the mint these numbers somehow?

mfoltzgoogle: No we didn't invent the numbers :-)

In public key certificates the algorithms and parameters are specified using Object IDs that are defined by IETF specs.  RFC 5480 defines the OIDs for ECDSA public keys and RFC 5758 defines them for ECDSA signature schemes.

These OIDs end up being encoded to binary using the ASN.1 encoding scheme and embedded in the certificate (along with the actual public key parameters which vary key-by-key).

My interpretation of the feedback from @sleevi in #280 was that to be crystal clear we should put the actual binary representations of these OIDs in the spec as DER (i.e., hexadecimal).  But this DER form is not in the RFCs, you have to do that conversion yourself ([tool here](https://misc.daniel-marschall.de/asn.1/oid-converter/online.php)) or parse an actual DER-encoded certificate.

I updated this PR to include both the object names, the OIDs, and the DER representations of the OIDs.  Does that make things clearer?

Since we may drop P384 from the supported certificates in PR #297 anyway,  I didn't fill in the row for P384 yet.  We can land #297 then update this PR accordingly.

index.bs
```diff
@@ -387,8 +388,25 @@ The [=agent certificate=] must have the following characteristics:
...
+     <p>
+      Object Identifiers:
+      <ol>
+        <li>`1.2.840.10045.2.1` (ECC)</li>
+        <li>`1.2.840.10045.3.1.7` (ECDSA P256)</li>
+      </ol>
+      DER representation: `301306072a8648ce3d020106082a8648ce3d030107`
+      </p>
+  </td>
+  <td>`ecdsaWithSha256`
```
tidoust, line 401: I suppose the object name is a reference to the definition in RFC5758. The RFC seems to use hyphens for this one:

```suggestion
  <td>`ecdsa-with-SHA256`
```

index.bs
tidoust: Many thanks for the explanation.

As long the table is clear that it contains the DER representation of the object identifiers defined in the RFCs, as done in your latest update, the object names and the object identifiers may not be needed since they represent the exact same thing. They are useful for laymen like me to figure things out, but I don't think I'm the targeted audience for this part in any case. Up to you!

mfoltzgoogle:
I have gone ahead and filled in the fields for `secp384r1` and `ecdsa-with-SHA384`.  Please take a look to see if you have any additional comments.

mfoltzgoogle:
Tagging @tidoust in case he didn't see my last comment :)

tidoust (approved):

Issue #276, opened 2021-04-06
Agent Certificate has a circular dependency on itself


sleevi:
The value `fp` is defined as: https://github.com/w3c/openscreenprotocol/blob/5488c7b7cce9c9c64ba97348a8deb0c6c50eb9bb/index.bs#L285-L291

Agent Certificate then follows, with the definition at:
https://github.com/w3c/openscreenprotocol/blob/5488c7b7cce9c9c64ba97348a8deb0c6c50eb9bb/index.bs#L354-L417

The issue is that the certificate Serial Number field makes use of `fp`, as described at
https://github.com/w3c/openscreenprotocol/blob/5488c7b7cce9c9c64ba97348a8deb0c6c50eb9bb/index.bs#L385-L388

The problem is that `fp` is computed over the certificate, so one cannot compute `fp` apriori to include within the certificate, and any inclusion within the certificate would naturally change the derived fingerprint.

mfoltzgoogle:
We also need to fix the certificate subject name not to refer to the certificate fingerprint.

backkem:
Regarding the subject name, similarly replacing it with the serial number could look as follows:

1. Advertise the serial number as a `sn` mDNS TXT record.
1. use `<sn>._openscreen._udp` as the certificate Subject CommonName.
1. The dialing client can use the  `sn` TXT record to construct the `<sn>._openscreen._udp` to use in the server_name extension.

This keeps the 1:1 relation between server name and certificate. It don't think it impacts the certificate validation based on the fingerprint. That being said, I may have missed some of the original considerations put into the current spec. If someone can validate that, I'm happy to do the writeup and send a PR for review.

backkem:
> It seems "Issuer Name" also has problem. According to the spec, "Issuer Name" is set as the model-name from the agent-info message. However, the model-name information is exchanged during the metadata phase which happens after the connection. Is there something wrong with my understanding?

My interpretation of this is that it's the model name of the agent creating the certificate itself, meaning the model name as it is also added to the agent-info message; not the model name as is received from the remote agent. Can someone confirm/deny this? If my interpretation is correct, we may want to rephrase this to avoid any confusion.

mfoltzgoogle:
> Regarding the subject name, similarly replacing it with the serial number could look as follows:

That sounds like a great approach.  bakkem@, do you have the capacity to put up a proposed PR that implements that?

mfoltzgoogle:
[...]

PR #293, opened 2022-09-07
Add an algorithm for setting the agent certificate serial number

mfoltzgoogle:
This addresses Issue #276: Agent Certificate has a circular dependency on itself

We can't use the certificate fingerprint as the serial number, because the serial number is embedded in the certificate itself.

X.509 requires every certificate from the same issuer to have a unique (not necessarily consecutive) serial number.

This PR adds a simple algorithm to generate a unique per-agent serial number per certificate.   Even if an agent generates a new certificate once per hour, the counter would take almost 500,000 years to overflow :-P

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/openscreenprotocol/pull/293.html" title="Last updated on Jan 22, 2024, 7:45 PM UTC (a90f8df)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/openscreenprotocol/293/15ed7a5...a90f8df.html" title="Last updated on Jan 22, 2024, 7:45 PM UTC (a90f8df)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -55,7 +55,6 @@ url: https://tools.ietf.org/html/rfc8122#section-5; type: dfn; spec: RFC8122; te
 url: https://tools.ietf.org/html/rfc8122#section-5; type: dfn; spec: RFC8122; text: md5
 url: https://tools.ietf.org/html/rfc6381#section-3; type: dfn; spec: RFC6381; text: codecs parameter
 url: https://tools.ietf.org/html/rfc8610#section-3; type: dfn; spec: RFC8610; text: concise data definition language
-url: https://datatracker.ietf.org/doc/html/rfc4122#section-4.4; type: dfn; spec: RFC4122; text: version 4 uuid
 </pre>
 
 Introduction {#introduction}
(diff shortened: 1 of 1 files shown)


mfoltzgoogle:
Feel free to review now in advance of TPAC.

mfoltzgoogle:
baylesj@, do you mind taking a look?

backkem:
Is the same comment not relevant to the `Subject Name`? It is defined as `CN = <fp>._openscreen._udp`.

index.bs
```diff
@@ -375,6 +375,26 @@ The [=agent certificate=] must have the following characteristics:
...
+Let the <dfn>certificate serial number</dfn> be the result of the following steps:
+
+<ol>
+  <li>If the agent has never generated an agent certificate:
+  <ol>
+    <li>Let the <dfn>certificate serial number base</dfn> be a 32-bit
+        pseudorandom integer value.</il>
+    <li>Let the <dfn>certificate serial number counter</dfn> be a 32-bit
+        unsigned integer.</li>
+    <li>Assign the [=certificate serial number counter=] to 0.</li>
```
baylesj, line 387: Nit: the base is declared and assigned in a single step. For consistency, have both the base and counter either declared and assigned in one step each, or two steps.

baylesj (approved):
LGTM % feedback.

backkem:
The use of <fp> in the subject name is no issue? Is there a way to set this field after calculation of the fingerprint?

index.bs
mfoltzgoogle: Done

mfoltzgoogle:
> Is the same comment not relevant to the `Subject Name`? It is defined as `CN = <fp>._openscreen._udp`.

Yes, that is also an issue, thank you for pointing that out.  We won't be able to close Issue #276 until we address that.

mfoltzgoogle:
PTAL @baylesj, I will have to come up with a separate proposal to fix the subject name.

mfoltzgoogle:
Went ahead and merged this as comments have been addressed. Please let me know if there are other comments @baylesj

Issue #235, opened 2019-10-07
Update SPAKE2 section according to current IETF draft


mfoltzgoogle:
New draft: https://tools.ietf.org/html/draft-irtf-cfrg-spake2-09
Diff: https://tools.ietf.org/rfcdiff?url2=draft-irtf-cfrg-spake2-09.txt

At a minimum, update the links.  Also make sure spec language reflects the current description of the key exchange protocol described by the draft.

mfoltzgoogle:
This is on hold pending outcome of #242.

PR #294, opened 2022-09-08
Update for SPAKE2

mfoltzgoogle:
This PR addresses Issue #235: Update SPAKE2 section according to current IETF draft

The IETF draft has had several updates since the SPAKE2 section was first written.   I am taking this opportunity to align
the protocol with the way the algorithm is written in the current draft, as an explicit two-round process.  

By moving to a two-round process the `auth-spake2-needs-psk` message can be eliminated and replaced by having each agent explicitly state the PSK status in the `auth-spake2-handshake` message.  This leads to a cleaner and easier to understand protocol.

In addition the message fields and associated text are updated to align with the terms used in the IETF draft.

Finally it updates the transcript-mac field to 64 bytes to allow the use of SHA-512 in the HMAC.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -649,26 +649,24 @@ that may lead to authentication failures.
 Authentication with SPAKE2 {#authentication-with-spake2}
 --------------------------
 
-Issue(235): Update SPAKE2 section according to current IETF draft
-
 Issue(242): [Meta] Track CFRG PAKE competition outcome
 
 For all messages and objects defined in this section, see Appendix A for
 the full CDDL definitions.
 
 The default authentication method is
-\[SPAKE2](https://tools.ietf.org/html/draft-irtf-cfrg-spake2-08) with
+\[SPAKE2](https://tools.ietf.org/html/draft-irtf-cfrg-spake2-26) with
 the following cipher suite:
 
-1. Elliptic curve is [Curve25519](https://tools.ietf.org/html/rfc7748#page-4).
-2. Hash function is \[SHA-512](https://tools.ietf.org/html/rfc6234).
+1. Elliptic curve is \[edwards25519](https://tools.ietf.org/html/rfc7748#page-4).
+2. Hash function is \[SHA-256](https://tools.ietf.org/html/rfc6234).
 3. Key derivation function is \[HKDF](https://tools.ietf.org/html/rfc5869).
 4. Message authentication code is \[HMAC](https://tools.ietf.org/html/rfc2104).
-5. Password hash function is SHA-512.
+5. Password hash function is \[SHA-512](https://tools.ietf.org/html/rfc6234).
 
-Open Screen Protocol does not use a memory-hard hash function to hash PSKs
-with SPAKE2 and uses SHA-512 instead as the PSK is one-time use and
-is not stored in any form.
+Open Screen Protocol does not use a memory-hard hash function to hash PSKs with
+SPAKE2 and uses SHA-512 instead, as the PSK is one-time use and is not stored in
+any form.
 
 SPAKE2 provides explicit mutual authentication.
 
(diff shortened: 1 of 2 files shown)


mfoltzgoogle:
This should be ready for review.

messages_appendix.cddl
```diff
@@ -75,23 +75,24 @@ auth-initiation-token = (
...
+; type key 1003
 auth-spake2-confirmation = {
-  1 : bytes .size 32 ; transcript-mac
+  1 : bytes .size 64 ; transcript-mac
 }
 
-; type key 1005
+; type key 1004
 auth-status = {
   1 : auth-status-result ; result
```
baylesj, line 97: Where is auth-status-result defined? It's not in this CDDL file?

index.bs
```diff
@@ -679,42 +677,48 @@ phone, a keyboard or a TV remote control.
...
+it to the the user.  Once that is done, it sends an [=auth-spake2-handshake=]
+message to the PSK consumer with `psk-status` set to `psk-input` and the
+`public-value` field set to `pB`.
+
+Once an agent knows both `pA` and `pB` from [=auth-spake2-handshake=] messages,
+it computes and sends a [=auth-spake2-confirmation=] with the `transcript-mac`
+field set to `cA` (for Alice) or `cB` (for Bob) to the other agent.
+
+Once an agent receives an [=auth-spake2-confirmation=] message, it validates
+that message using the procedure in \[SPAKE2] and send an [=auth-status=]
```
baylesj, line 716: and send*s* an

or consider: "and then replies with"

index.bs
```diff
@@ -679,42 +677,48 @@ phone, a keyboard or a TV remote control.
...
+authenticated message to the other agent.  Any value of `result` other than
+`authenticated` means that authentication failed.
+
+Note that this message is merely informative as each agent independently
```
baylesj, line 720: Could this ever be more than just informative? E.g. should the agent wait for an auth-status message in case the other agent uses a different version of the protocol and gets a different result? Or is it effectively ignored?

messages_appendix.cddl
```diff
@@ -75,23 +75,24 @@ auth-initiation-token = (
...
-  1 : bytes .size 32 ; random
+  0: auth-spake2-psk-status ; psk-status
+  1: bytes ; public-value
 }
 
-; type key 1004
+; type key 1003
 auth-spake2-confirmation = {
-  1 : bytes .size 32 ; transcript-mac
+  1 : bytes .size 64 ; transcript-mac
```
baylesj, line 92: perhaps out of scope, but `transcript-mac` seems like an odd naming choice to me. MAC is the message authentication code algorithm right (as opposed to the more globally used media access control address). `transcript-mac` then corresponds to `Ke` or `Ka` right?

Maybe name this `mac-key` or `confirmation-key` then?

index.bs
```diff
@@ -711,31 +711,30 @@ QR-code and a numeric PSK), they should be for the same PSK.  If they were
...
 3. Key derivation function is \[HKDF](https://tools.ietf.org/html/rfc5869).
 4. Message authentication code is \[HMAC](https://tools.ietf.org/html/rfc2104).
-5. Password hash function is SHA-512.
+5. Password hash function is \[SHA-512](https://tools.ietf.org/html/rfc6234).
 
-Open Screen Protocol does not use a memory-hard hash function to hash PSKs
-with SPAKE2 and uses SHA-512 instead as the PSK is one-time use and
-is not stored in any form.
+Open Screen Protocol does not use a memory-hard hash function to hash PSKs with
+SPAKE2 and uses SHA-512 instead, as the PSK is one-time use and is not stored in
```
baylesj, line 736: Nit: you updated the hash function to SHA-256 so this comment is now out of date.

messages_appendix.cddl
```diff
@@ -75,23 +75,24 @@ auth-initiation-token = (
...
-  auth-initiation-token
-}
+auth-spake2-psk-status = &(
+ psk-needs-presentation: 0
+ psk-shown: 1
+ psk-input: 2
+)
 
-; type key 1003
+; type key 1005
```
baylesj, line 84: Should the order of these types be changed now that the type keys are not organized by value?

baylesj (approved):
LGTM % feedback.

index.bs
mfoltzgoogle: I don't think I did, the hash function for SPAKE2 is set to `SHA-512` in point #5 above.

messages_appendix.cddl
mfoltzgoogle: Done

messages_appendix.cddl
mfoltzgoogle: It's defined below, I moved it up right before where it is used here.

mfoltzgoogle:
PTAL @baylesj

Issue #233, opened 2019-09-23
Use the same codec names as the media APIs.


pthatcherg:
MediaCapabilities (and other media APIs) uses a content type of "video/foo; codecs=A.B.C". We should probably use the "A.B.C" instead of what we have currently. For example, instead of "h264", use "avc1.4200" or instead of "vp9", "vp09.00.10.08". It's a little crazy and I can't find any RFCs for these things, but here is where Chromium parses them:

https://cs.chromium.org/chromium/src/media/base/mime_util_internal.cc?g=0&l=820

mfoltzgoogle:
Is this what you are looking for?  It covers some, but not all of the MPEG codecs.

https://tools.ietf.org/html/rfc6381

There may be an equivalent RFC for the VP*/AV1 codecs.

The media APIs don't make any explicit reference to any RFCs.  I have no idea what implementers are supposed to do, other than copy each other's source code :-P

mfoltzgoogle:
Spoke with Dan Sanders, there are other RFCs for the webm container we can refer to as well.

He noted that enumeration of codec strings for any given codec family is a hard problem because of the incredibly rich syntax for codecs - so we should avoid that in our protocol design.

mfoltzgoogle:
Take a look at these specs for VPX and AV1 and you'll see what I mean:

https://storage.googleapis.com/downloads.webmproject.org/docs/vp9/vp-codec-iso-media-file-format-binding-20160516-draft.pdf

https://aomediacodec.github.io/av1-isobmff/#codecsparam

mfoltzgoogle:
It sounds like there wasn't an explicit action recorded in the last F2F minutes [1], but for the record, we discussed consulting with the Media WG folks working on Media Capabilities and related APIs to get feedback on this issue.

[1] https://www.w3.org/2021/02/23-webscreens-minutes.html#t02

chrisn:
The minutes from the joint meeting with Media WG are [here](https://www.w3.org/2022/01/11-mediawg-minutes.html#t02). Summary:

* We can use references in the WebCodecs registry for codec string details: https://www.w3.org/TR/webcodecs-codec-registry/

mfoltzgoogle:
Want to discuss briefly and propose putting up a pull request to update the specification to explicitly use the WebCodecs registry to resolve this issue.

mfoltzgoogle:
[...]

PR #299, opened 2022-09-09
Reference WebCodecs registry for codec names.

mfoltzgoogle:
Addresses Issue #233: Use the same codec names as the media APIs.

Per the discussion in https://www.w3.org/2022/06/08-webscreens-minutes.html#t08, this updates the spec to refer explicitly to the WebCodecs registry to define codec strings for codecs in that registry.

It adds fallback language for additional codecs not defined in that registry, but if it is not sufficient to define the strings for additional codecs I would prefer to just remove the fallback language from this PR and address that separately.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/openscreenprotocol/pull/299.html" title="Last updated on Sep 9, 2022, 10:23 PM UTC (e104c11)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/openscreenprotocol/299/e8e728c...e104c11.html" title="Last updated on Sep 9, 2022, 10:23 PM UTC (e104c11)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1647,14 +1647,17 @@ The format type is used as the basis for audio and video capabilities.
 Formats are composed of the following fields:
 
 : name (required)
-:: The name of the codec.   This must be a single-codec RFC 6381 [=codecs parameter=].
-
-Issue(233): Use the same codec names as the media APIs.
+:: A fully qualified codec string listed in the [[WEBCODECS-CODEC-REGISTRY]] and further
+     specified by the codec-specific registrations referenced in that registry.
 
 : parameters (required)
 :: A list of (key, value) parameters that can be used to describe
     properties of a specific format, and not shared by other formats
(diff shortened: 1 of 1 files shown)


tidoust (approved):

chrisn (approved):

mfoltzgoogle:
This was discussed at TPAC 2022 [1] and there weren't any specific objections to going this route to name codecs. 

An issue was raised about how agents can use capabilities to request a lower-resolution stream, because they are not capable of displaying the stream at original/higher resolution.  I'll file a separate issue for that; I don't think the exact details of how codecs are named in capabilities is directly related, since display and stream resolution are conveyed through separate, additional fields in the protocol.

[1] https://www.w3.org/2022/09/16-webscreens-minutes.html#t07

Issue #194, opened 2019-08-26
[Remote Playback] Capabilities for HDR rendering and display


mfoltzgoogle:
This issue is tracking requirements to allow one agent to determine another's capabilities to render and display HDR media, both for streaming and remote playback.

At the Berlin F2F [1], it was noted that HDR capabilities are under discussion in the Media WG, and this issue is tracking the work being done there to align the protocol with the data model that will be used:

- [x] https://github.com/w3c/media-capabilities/issues/118: Add Support for Querying HDR Decode and Render Capabilities
- [x] https://github.com/w3c/media-capabilities/issues/119: Add Support for Querying HDR & 4K Video Display Capabilities

It looks likely that this topic will be on the agenda of the Media WG at TPAC 2019.

mfoltzgoogle:
Ref: https://www.w3.org/2019/09/15-webscreens-minutes.html#x16

w3c/media-capabilities#118 has been closed, but w3c/media-capabilities#119 remains open.  I think it would make sense to land a PR to align with #118.

anssiko:
>ACTION: @mfoltzgoogle to propose a PR to align OSP with Media Capabilities issue https://github.com/w3c/media-capabilities/issues/118

via https://www.w3.org/2020/10/20-webscreens-minutes.html#a04

anssiko:
w3c/media-capabilities#119 was closed and spawned two issues:

- [ ] `Screen.video.hdrSupported` https://github.com/w3c/media-capabilities/issues/135
- [ ]  CSS length units in `video-*` media features https://github.com/w3c/csswg-drafts/issues/5044

mfoltzgoogle:
It sounds like there wasn't an explicit action recorded in the last F2F minutes [1], but for the record, we discussed consulting with the Media WG folks working on Media Capabilities and related APIs to get feedback on PR #225, to resolve the first of these two issues (the render/decoding capabilities).

[1] https://www.w3.org/2021/02/23-webscreens-minutes.html#t02

chrisn:
Please let me know if you'd like to bring this to a Media WG meeting, happy to put this on the agenda.

anssiko:
@chrisn given https://github.com/w3c/media-capabilities/issues/135 and w3c/csswg-drafts#5044 haven't received attention recently I'd appreciate that. Please drop meeting logistics pointer to this issue so interested people can join.

chrisn:
[...]

PR #300, opened 2022-09-09
Add transfer-functions and hdr-metadata to video-capabilities.

mfoltzgoogle:
Addresses Issue #194: [Remote Playback] Capabilities for HDR rendering and display

This adds two fields to the `receive-video-capability` message that allows the receiver to list the HDR metadata types and transfer functions that the media receiver can accept.

The values are used by reference from the corresponding enums in the Media Capabilities API.  This was discussed, briefly, in the minutes of the joint SSWG/Media WG meeting:

https://www.w3.org/2022/01/11-mediawg-minutes.html#t01

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/openscreenprotocol/pull/300.html" title="Last updated on Oct 31, 2022, 5:38 PM UTC (cd9280d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/openscreenprotocol/300/37cef31...cd9280d.html" title="Last updated on Oct 31, 2022, 5:38 PM UTC (cd9280d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1716,7 +1716,7 @@ following additional fields:
 
     The `transfer-function` field must be a valid
     [[MEDIA-CAPABILITIES#transferfunction|TransferFunction]]
-    and the `hdr-metadata` must be a valid
+    and the `hdr-metadata` field must be a valid
     [[MEDIA-CAPABILITIES#hdrmetadatatype|HdrMetadataType]], both defined in the
     [[!MEDIA-CAPABILITIES|Media Capabilities]] API.
 
@@ -1725,6 +1725,7 @@ following additional fields:
     without any associated metadata.  (This is the case, for example, with the
     "hlg" `transfer-function`.)
 
+    The media receiver should ignore duplicate entries in `hdr-formats.`
     If no `hdr-formats` are listed, then the media reciever cannot decode any
     HDR formats.
 
--- a/index.bs
+++ b/index.bs
@@ -1694,14 +1694,25 @@ following additional fields:
     display could return this value as 1.6 to indicate its preferred content
     scaling. Default is none.
 
-Issue(194): Expose capabilities for decoding HDR transfer functions and metadata formats.
-
 : color-gamuts (optional)
 :: An optional field indicating what color spaces can be decoded and rendered by
     the media receiver.  The media sender may use these values to determine how
     to encode video. Valid values correspond to [[MEDIA-CAPABILITIES#colorgamut|ColorGamut]]
     in the [[!MEDIA-CAPABILITIES|Media Capabilities]] API.  The default value is
     a list with the single entry "srgb".
+
+: transfer-functions (optional)
+:: An optional field indicating what HDR transfer functions can be decoded and rendered
+    by the media receiver.  Valid values correspond to
+    [[MEDIA-CAPABILITIES#transferfunction|TransferFunction]] in the
+    [[!MEDIA-CAPABILITIES|Media Capabilities]] API.  The default value is a list with the
+    single entry "srgb".
+
+: hdr-metadata (optional)
+:: An optional field indicating what HDR metadata types can be decoded and rendered by
+    by the media receiver.  Valid values correspond to
+    [[MEDIA-CAPABILITIES#hdrmetadatatype|HdrMetadataType]] in the
+    [[!MEDIA-CAPABILITIES|Media Capabilities]] API.  The default value is an empty list.
 
 : native-resolutions (optional)
 :: An optional field indicating what video-resolutions the media receiver supports and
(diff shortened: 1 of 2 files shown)


mfoltzgoogle:
One question that could use further input is whether the two new values are independent, i.e. if you have:

```
transfer-functions: [  "p3", "rec2020"  ]
hdr-metadata: [ "smpteSt2086", "smpteSt2094-10" ]
```
Is it always true that all four combinations are possible?  If not, then we might have to adjust the data model to be a list of `transfer-function` and `hdr-metadata` pairs.

Probably requires looking at some HDR displays to see what the constraints are in the real world.

mfoltzgoogle:
After reviewing some of the relevant documentation around the three linked fields `color-gamuts`, `transfer-functions` and `hdr-metadata`, here is what I think makes sense for the protocol.

- The color gamut values are inclusive by definition: `rec2020` includes `p3`, which includes `srgb`.  We could change `color-gamuts` from a list to a single field, as long as we made it clear that the agent receiving the `color-gamut` applied the inclusion logic.   I will file an issue for this, but I don't feel strongly that this requires a change to the protocol.

- Some `hdr-metadata` and `transfer-functions` values don't make sense.  For example HLG does not require a separate metadata format, so a media receiver that understood both e.g. HDR10+ and HLG via one codec would have to report them as two separate encoding types with all other fields duplicated:

Encoding 1:
```
<repeat everything else>
hdr-metadata: []
transfer-functions: ['hlg']
<repeat everything else>
```

Encoding 2:
```
<repeat everything else>
hdr-metadata: ['smpteSt2094-40']
transfer-functions: ['pq']
<repeat everything else>
```

For this reason I think we should report the HDR capabilities as a list of pairs, so this can be conveyed in one encoding:

```
<snip>
hdr-formats: [
  { transfer-function: 'hlg' },
  { transfer-function: 'pq', hdr-metadata: 'smpteSt2094-40' }
]
<snip>
```

meaning that encoding could convey either HLG or HDR10+.  For completeness, here is what it would look like with `smpteSt2094-10` included, which can support either HLG or PQ [1].

```
<snip>
hdr-formats: [
  { transfer-function: 'hlg' },  // HLG with no metadata
  { transfer-function: 'hlg', hdr-metadata: 'smpteSt2094-10' }  // 2094-10 with HLG. [2]
  { transfer-function: 'pq', hdr-metadata: 'smpteSt2094-10' }   // 2094-10 with PQ
]
<snip>
```

Note, per the issue linked below, Dolby has registered a number of MIME types to represent their proprietary HDR metadata encodings, so remote playback capability detection for these formats would be handled by the `remote-playback-source` message, which accepts full MIME types.

[1] https://github.com/w3c/media-capabilities/issues/136
[2] Unclear if if this happens in practice as HLG support may be a proprietary profile of Dolby Vision and not part of smpteSt2094-10.

mfoltzgoogle:
@chrisn, we have resolved the issues preventing PRs from landing on the OSP spec.  Feel free to take a look.

chrisn:
Thank you, somehow I missed your previous comment here, I'll take a look.

chrisn:
> We could change color-gamuts from a list to a single field, as long as we made it clear that the agent receiving the color-gamut applied the inclusion logic. I will file an issue for this, but I don't feel strongly that this requires a change to the protocol.

I agree.

> For this reason I think we should report the HDR capabilities as a list of pairs

This sounds good.

> Dolby has registered a number of MIME types to represent their proprietary HDR metadata encodings, so remote playback capability detection for these formats would be handled by the remote-playback-source message, which accepts full MIME types.

I see. If we need to we can follow this up with them via the new [A/V Media Formats CG](https://github.com/w3c/avmedia-formats-cg), which started at TPAC 2022.

mfoltzgoogle:
> > We could change color-gamuts from a list to a single field, as long as we made it clear that the agent receiving the color-gamut applied the inclusion logic. I will file an issue for this, but I don't feel strongly that this requires a change to the protocol.
> 
> I agree.

Filed https://github.com/w3c/openscreenprotocol/issues/305

mfoltzgoogle:
> > For this reason I think we should report the HDR capabilities as a list of pairs
> 
> This sounds good.

Ok, updated PR to reflect this change.

mfoltzgoogle:
Feel free to review the current PR @chrisn with the changes I proposed, however, the spec generation is currently broken again so you won't be able to see an HTML preview yet. ☹️

chrisn:
Thanks. I've asked a colleague for advice on this part:

> [2] Unclear if this happens in practice as HLG support may be a proprietary profile of Dolby Vision and not part of smpteSt2094-10.

mfoltzgoogle:
Hi @chrisn, did you have any additional feedback on the PR? 

Regarding:

> [2] Unclear if this happens in practice as HLG support may be a proprietary profile of Dolby Vision and not part of smpteSt2094-10.

Would definitely appreciate any feedback on this.  However, the protocol proposed in the PR should support this use case if it is necessary.

chrisn (approved):
This is good, thanks!

mfoltzgoogle:
Thanks!

PR #311, opened 2023-09-08
Convert color-gamut to be a single value.

### LINKED ISSUE #305: `color-gamuts` could be a single value and not a list
Following up from https://github.com/w3c/openscreenprotocol/pull/300#issuecomment-1264085747

> The color gamut values are inclusive by definition: rec2020 includes p3, which includes srgb. We could change color-gamuts from a list to a single field, as long as we made it clear that the agent receiving the color-gamut applied the inclusion logic. I will file an issue for this, but I don't feel strongly that this requires a change to the protocol.

The protocol change proposed would be to replace the `color-gamuts` field in the `receive-video-capability` dictionary from a list of values, to a single value.

mfoltzgoogle:
Fixes #305.  This converts color-gamuts to be a single value, not a list, as support for a wider gamut implies support for narrower gamuts.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/openscreenprotocol/pull/311.html" title="Last updated on Oct 13, 2023, 10:21 PM UTC (7cd67df)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/openscreenprotocol/311/212162c...7cd67df.html" title="Last updated on Oct 13, 2023, 10:21 PM UTC (7cd67df)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1770,12 +1770,15 @@ following additional fields:
     display could return this value as 1.6 to indicate its preferred content
     scaling. Default is none.
 
-: color-gamuts (optional)
+: color-gamut (optional)
 :: An optional field indicating what color spaces can be decoded and rendered by
-    the media receiver.  The media sender may use these values to determine how
+    the media receiver.  The media sender may use this value to determine how
     to encode video. Valid values correspond to [[MEDIA-CAPABILITIES#colorgamut|ColorGamut]]
     in the [[!MEDIA-CAPABILITIES|Media Capabilities]] API.  The default value is
-    a list with the single entry "srgb".
+    "srgb".
+
+NOTE: Support for "p3" implies support for "srgb", and support for "rec2020"
+implies support for "p3" and "srgb".
 
 : hdr-formats (optional)
 :: An optional field indicating what HDR transfer functions and metadata formats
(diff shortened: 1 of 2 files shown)


mfoltzgoogle:
@chrisn, I recall you helped in #300 which declares color capabilities in the protocol, and this is a follow-up to that PR.

index.bs
```diff
@@ -1770,12 +1770,15 @@ following additional fields:
     display could return this value as 1.6 to indicate its preferred content
     scaling. Default is none.
 
-: color-gamuts (optional)
+: color-gamut (optional)
 :: An optional field indicating what color spaces can be decoded and rendered by
```
baylesj, line 1774: I would consider rephrasing this to make it clear that it is the widest color space that the client supports. 

```
An optional field indicating the widest color space that can be decoded and rendered by
the media receiver. The media sender may use this value to determine how to encode
video, and may assume that all narrower color spaces are also supported.
```

baylesj: I know you do it below with the `NOTE:` section so feel free to ignore this advice if you prefer that approach instead.

baylesj (approved):
LGTM % feedback.

chrisn (approved):

index.bs

mfoltzgoogle:
baylesj@: I merged based on your LGTM; note you can explicitly approve PRs using the dropdown in the GitHub UI.

Issue #265, opened 2021-03-03
Should media resource URLs be optional?


mfoltzgoogle:
If a controller has a media element using Media Source Extensions, there may be no media resource URL at all; the application will attach a MediaSource directly to the media element by setting its `srcObject`.

We need to allow this scenario in the definitions of the `remote-playback-*` messages, and also clarify how to handle transitions between `src` (list of media resources) and `srcObject` in the controlling media element.

Issue #223, opened 2019-09-11
Codec switching when remoting.


mfoltzgoogle:
From https://github.com/webscreens/openscreenprotocol/pull/172#issuecomment-529734137 and the associated review thread, we identified a need for a streaming session to switch codecs; for example when a media element being remoted changes the type of its MSE source.  Other use cases are if the source material changes resolution (which might require a more efficient codec) or if network conditions allow upgrading from lossy to lossless codecs.

@pthatcherg suggested we could have the receiver enumerate codecs at the start of the session and allow the sender to switch mid-session.

mfoltzgoogle:
https://www.w3.org/2019/09/15-webscreens-minutes.html#x18

> RESOLVED: We'll try 1. Receiver capabilities that allow the streaming sender to switch codecs, plus 2. An API on RemotePlayback that allow JS to query post-prompt for capabilities like the MediaCapabilities API does (issue #223)

Item #1 can be accommodated strictly within the OSP specification, but #2 will require spinning off an issue for the Remote Playback API and figuring out how that fits into the scope of work for it.

anssiko:
>ACTION: @mfoltzgoogle to add "needs research" on issue #223 for second point

via https://www.w3.org/2020/10/20-webscreens-minutes.html#a03

mfoltzgoogle:
Based on the discussion linked above, it sounds like there is interest in exposing the capabilities of the receiving agent to script while remote playback is active, so that the site can customize the media that is sent (for example, choosing a more appropriate codec, color space, audio track, etc.) .  That API would need to be defined first in the Remote Playback API, to determine if any changes are needed in the protocol.

Separately from the issue of exposing capabilities, if the site changes the media source during remote playback, the protocol supports this by sending a `remote-playback-modify-request` with the `remote-playback-controls.source-url` field set to a new value.  However, we are not passing the `extended-mime-type` like we do with the `remote-playback-start-request`, so that is something that can be fixed in the spec.

If the remote playback session is remoting (i.e. sending encoded media, not a URL), then there isn't currently a way for the sender to offer new encodings during an existing session; only the receiver can make a request to modify the encodings sent during a session.  As a workaround, the sender can terminate the current session, and start a new session by offering new encodings.   I think this is good enough for now, but once we have a working implementation we can determine if this causes quality issues (i.e., a gap in playback before the new encoding starts to play out).

PR #313, opened 2023-09-10
Address remote playback issues.

mfoltzgoogle:
This PR addresses Issues #223 and #265

- It changes the remote playback state / controls to include the full media resource (URL + mime type), so that the sender can change the media resource after remote playback has started.
- It describes what the sender needs to do to change the encoding for a stream after a streaming session has started.
- It makes the source optional for a `remote-playback-start-request` if the remote playback begins with media remoting (streaming session)

Note, I'm wondering if `remote-playback-controls` should have a list of sources like the `remote-playback-start-request`...

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1100,8 +1100,6 @@ particular URL or set of URLs, the controller may send a
 :: A list of [=media resources=], the same as specified in the
     [=remote-playback-start-request=] message.  Must not be empty.
 
-Issue(146): Remote Playback HTTP headers.
-
 : headers
 :: headers that the receiver should use to fetch the
     urls.  For example,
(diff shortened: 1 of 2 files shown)


mfoltzgoogle:
@baylesj Do you mind reviewing this one as well?
@muyao-xu as this impacts the Remote Playback API.

index.bs
```diff
@@ -1162,10 +1160,10 @@ requirements for a remote-playback-id.
 : sources
 :: The [=media resources=] that the controller has selected for playback
     on the receiver.  Each source must include a <{source/src|source URL}>
-    and should include an <{source/type|extended MIME type}>.
+    and should include an <{source/type|extended MIME type}>.  If empty,
```
baylesj, line 1163: If it's allowed to be empty, isn't this a "may include"?

index.bs
```diff
@@ -1162,10 +1160,10 @@ requirements for a remote-playback-id.
 : sources
```
baylesj, line 1160: This is now optional right?

baylesj (approved):
LGTM % feedback.

index.bs
mfoltzgoogle: Yes, it can be empty, but if the media element has one it should be included in the `remote-playback-start-request`.  Added some text to clarify.

mfoltzgoogle: Sorry, thought that comment was about the mime type :)

index.bs

mfoltzgoogle:
PTAL

baylesj (approved):
LGTM.

Issue #254, opened 2021-02-27
[Capabilities] Finish the capabilities table.


mfoltzgoogle:
The table that maps Open Screen [capabilities](../capabilities.md) to supported message types has no supported messages for the Streaming Sender.   We should complete the table and double check the entries.

As a bonus, a second table matching of message type IDs to message names would make the table easier to read.

PR #314, opened 2023-09-10
Update capabilities table.

mfoltzgoogle:
This addresses #254 by assigning remaining message types in the protocol to capabilities in the table.

A new capability `receive-data` is added to the table since there was no existing capability to receive message type 24 (data frames).

`video-request-format` was removed from the list of message definitions because there was no reference to it in the spec.


mfoltzgoogle:
Merging as this is editorial, except for the addition of the `receive-data` capability.

Issue #327, opened 2024-02-27
Representation of time in remote-playback-state


mfoltzgoogle:
There are a few things that would simplify the time related fields in `remote-playback-state`.

Currently they all use `media-time` which is the internal clock used by the media renderer to synchronize audio/video frames.  However HTML uses float values in the API to control media playback.

- `epoch` should be milliseconds since the epoch (positive or negative).  It could be a float64 or an int64.
- `duration` and `currentTime` could be a float64 to align with HTML.  It depends on whether it's the responsibility of the remote playback controller or the remote playback renderer to convert between `media-time` and seconds on the media timeline.
- Similar case for the media time ranges - they should follow whatever is done for `currentTime`.
- I think we should use `null` to represent an unknown value instead of having a separate type field.  We can't just omit the field because that means "no change from previous state update."

As part of this PR the spec should propose how time values should be converted to/from `media-time` for use in the protocol.

PR #328, opened 2024-02-28
Update media time usage for remote playback protocol.

mfoltzgoogle:
Addresses and closes Issue #327: Representation of time in remote-playback-state

The current spec uses the per-frame media synchronization timestamps to control and update the media playback state and handle other time specific fields.  This has a few issues:

1. The media start date is defined as milliseconds since the epoch (1/1/1970 00:00), not anything related to a specific media playback.
2. Other media times are all defined as floating point seconds on the media timeline. 
3. Only media playbacks that use streaming (instead of just sending a URL) will use the media synchronization time at all.

Therefore, to align with HTML, we just convert time fields to correspond to the types that are exposed in script on media elements, so that agents don't have to translate to/from media synchronization time.  (For cases where conversion is necessary, I added an appendix with a formula to do that.)

The protocol also used "unknown" to represent an unknown value, which does not have meaning in CBOR or CDDL. This PR also updates the protocol to use CBOR `null` instead which is a primitive value in CBOR.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/openscreenprotocol/pull/328.html" title="Last updated on Feb 28, 2024, 11:55 PM UTC (ffea304)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/openscreenprotocol/328/d338b24...ffea304.html" title="Last updated on Feb 28, 2024, 11:55 PM UTC (ffea304)">Diff</a>


index.bs
```diff
@@ -2898,3 +2902,18 @@ To decode a PSK `P` given a QR code, follow these steps:
...
+To convert between a media synchronization timestamp for a given audio or video
+frame and a media timeline value, the following formula can be used:
+
+```
+media-timeline-value = media-zero-time + (value / scale)
+```
+
+Where:
+- `media-zero-time` is the origin of the [=media timeline=] as defined in HTML.
+- `value` and `scale` are the values passed in the `sync-time` field of the
```
baylesj, line 2918: Might be pendantic, but might be worth suggesting how the math should be calculated since value and scale are both `uint` types and media-timeline-value is presumably a `float64`. 

My guess is that both value and scale should be converted to `float64` before division occurs in order to avoid rounding?

messages_appendix.cddl
```diff
@@ -422,16 +426,16 @@ text-track-mode = &(
...
   1: text ; id
-  2: media-time-range ; range
+  2: media-timeline-range ; range
   3: text ; text
 }
 
-media-time-range = [
-  start: media-time,
-  end: media-time
+media-timeline-range = [
```
baylesj, line 433: Nit: consider declaring above next to `media-timeline`, especially since it is used above.

baylesj (approved):

messages_appendix.cddl
mfoltzgoogle: Done

index.bs
mfoltzgoogle: Added specific details about how `media-timeline-value` and the computation should be done, and also what to do in event of an overflow (pin to the max value).

PR #329, opened 2024-04-24
Add entire flow chart.

wangw-1991:
Add an entire flow chart to make the entire process clearer.

***
<a href="https://pr-preview.s3.amazonaws.com/wangw-1991/openscreenprotocol/pull/329.html" title="Last updated on Apr 25, 2024, 1:22 AM UTC (67e03b4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/openscreenprotocol/329/d338b24...wangw-1991:67e03b4.html" title="Last updated on Apr 25, 2024, 1:22 AM UTC (67e03b4)">Diff</a>


wangw-1991:
Hi, @mfoltzgoogle. Can you help review this? Thanks.

index.bs
```diff
@@ -2898,3 +2903,7 @@ To decode a PSK `P` given a QR code, follow these steps:
 
 1. Obtain the string `N` by decoding the QR code.
 2. Parse `N` as a hexadecimal number to obtain `P`.
+
+Appendix D: Entire Flow Chart {#appendix-d}
+=============================
+<img src="./images/entire_flow_chart.svg" alt="entire flow chart">
```
anssiko, line 2909:
```suggestion
<img src="./images/entire_flow_chart.svg" alt="Entire flow chart" style="width: 100%; height: 100%">
```

This will fix the following markup validator error:
`error: Bad value “100%” for attribute “height” on element “img”: Expected a digit but saw “%” instead.`

anssiko (changes requested):

anssiko:
@wangw-1991 thank you for this great contribution!

My expectation is this diagram is aligned with the normative text in the specification and is considered non-normative similarly to examples and notes given it is an appendix and a diagram.  PTAL @mfoltzgoogle for alignment with the normative text in the specification.

@tidoust please let us know if we should explicitly mark this as non-normative.

@backkem FYI

index.bs
wangw-1991: Thanks anssi for the suggestion, I have changed this.

anssiko (approved):

chrisn:
This looks really good! It would be nice to see diagrams to show the metadata and application message exchanges too, as a suggestion.

wangw-1991:
> This looks really good! It would be nice to see diagrams to show the metadata and application message exchanges too, as a suggestion.

Thanks for the suggestion. I don't show the metadata and application message exchanges here because there are many different types of metadata and application data. The use of abstract representation here means it is applicable to all types and it can also simplify the chart. Do you think it is reasonable?

chrisn:
I think this diagram is really good as is, so no suggestion to change it. Additional diagrams, if we make them, could be specific to different use cases or scenarios, e.g., establishing a streaming session.

index.bs
```diff
@@ -2898,3 +2903,7 @@ To decode a PSK `P` given a QR code, follow these steps:
 
 1. Obtain the string `N` by decoding the QR code.
 2. Parse `N` as a hexadecimal number to obtain `P`.
+
+Appendix D: Entire Flow Chart {#appendix-d}
+=============================
+<img src="./images/entire_flow_chart.svg" alt="Entire flow chart" style="width: 100%; height: 100%">
```
tidoust, line 2909: It seems a good idea to flag the section as informative to clarify that, in case of discrepancy, the normative prose takes precedence. Bikeshed does not have a magic keyword to do that, so section needs to be wrapped in a `<section>` and text added explicitly.

I also propose to provide a more complete `alt` text that describes the image to meet [HTML requirements](https://html.spec.whatwg.org/multipage/images.html#alt).

Two other changes:
- I'm not sure what `height: 100%` changes. I dropped it. Feel free to reintroduce if I missed something!
- The `no-autosize` attribute tells Bikeshed not to try to compute the dimensions of the image (which it cannot because it's an SVG file).

```suggestion
<section class="informative">
Appendix D: Entire Flow Chart {#appendix-d}
=============================
*This section is non-normative.*

<img no-autosize src="./images/entire_flow_chart.svg" alt="Before a listening agent (client) and an advertising agent (server) may exchange application messages, they first need to discover each other through an mDNS exchange, establish a QUIC connection through ClientHello and ServerHello exchanges and sharing of certificates, and run an SPAKE2 authentication handshake and confirmation. They may also exchange messages to discover metadata at any time after a QUIC connection has been established." style="width: 100%">
</section>
```

anssiko: Thanks @tidoust! @wangw-1991 to integrate and @mfoltzgoogle to review.

mfoltzgoogle (approved):

mfoltzgoogle:
LGTM with Francois' suggestions applied.  Thank you again for this contribution!

index.bs
wangw-1991: Thanks @tidoust for the suggestions. Can you help merge this?

mfoltzgoogle (approved):

Issue #275, opened 2021-04-06
TLS SNI requirement is incompatible with TLS SNI definition


sleevi:
Section 4.1 contains the following language: https://github.com/w3c/openscreenprotocol/blob/5488c7b7cce9c9c64ba97348a8deb0c6c50eb9bb/index.bs#L343-L345

This may be incompatible with the definition of SNI from [RFC 6066, Section 3](https://tools.ietf.org/html/rfc6066#section-3), which normatively references [RFC 5890](https://tools.ietf.org/html/rfc5890), which has LDH labels (of which A-Labels are a subset), expanded on in [Section 2.3.1](https://tools.ietf.org/html/rfc5890#section-2.3.1)

The issue here is that underscore, as used by these records, is not part of the Preferred Name Syntax (which only allows Letters, Digits, and Hyphens). This is also called out within [RFC 8499](https://tools.ietf.org/html/rfc8499#section-2), which states:

>      This term and its equivalent, "hostname", have been
>       widely used but are not defined in [RFC1034], [RFC1035],
>       [RFC1123], or [RFC2181].  The DNS was originally deployed into the
>       Host Tables environment as outlined in [RFC952], and it is likely
>       that the term followed informally from the definition there.  Over
>       time, the definition seems to have shifted.  "Host name" is often
>       meant to be a domain name that follows the rules in Section 3.5 of
>       [RFC1034], which is also called the "preferred name syntax".  (In
>       that syntax, every character in each label is a letter, a digit,
>       or a hyphen).  Note that any label in a domain name can contain
>       any octet value; hostnames are generally considered to be domain
>       names where every label follows the rules in the "preferred name
>       syntax", with the amendment that labels can start with ASCII
>       digits (this amendment comes from Section 2.1 of [RFC1123]).

It's not obvious why SNI needs to be supported for this use case, so I'm not sure where to offer productive suggestions (e.g. as an Errata/attempt to redefine SNI to unambiguously allow arbitrary _domain_ names, and not just _host_ names, or some other TLS approach).

Could you clarify why the server needs to know the expected fingerprint, given that it will have advertised that via mDNS? If a certificate rotation event occurs, presumably, the mDNS update flow can handle this?

mfoltzgoogle:
Proposal by David Schinazi in code review:

[...]

PR #332, opened 2024-04-30
Use certificate serial number in Subject Name

### LINKED ISSUE #276: Agent Certificate has a circular dependency on itself

backkem:
This resolves the circular dependency in agent certificate fingerprint.

Resolves #276

***
<a href="https://pr-preview.s3.amazonaws.com/backkem/openscreenprotocol/pull/332.html" title="Last updated on May 16, 2024, 7:33 PM UTC (ecfce2e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/openscreenprotocol/332/362a2fd...backkem:ecfce2e.html" title="Last updated on May 16, 2024, 7:33 PM UTC (ecfce2e)">Diff</a>


anssiko (approved):
LGTM when the indentation issue has been fixed.

anssiko:
@mfoltzgoogle PTAL?

index.bs
```diff
@@ -359,8 +384,12 @@ refuse incoming connections that lack these parameters.
 
 * The [[!RFC7301|ALPN]] used must be "osp".
 * The [[!RFC6066|server_name extension]] must be set to the following `host_name`:
```
mfoltzgoogle, line 386: Let's leave updates to the server name extension out of this PR as here is an active discussion in #275 on this.

backkem: I'm trying to understand if it's feasible to consider these independently. As far as I can tell: if SNI is used to determine the server's identity, then this value should also be used to validate the server identity of the TLS certificate (`dNSName` or `CN`).

Next to that, #275 seems to have an issue with the `_openscreen._udp.local` post-fix rather than the `<fp>` or `<sn>` prefix. As these names are used for connection and no longer for mDNS discovery it also seems feasible to just use `<sn>.local` as server identity.

References: [RFC2818](https://www.rfc-editor.org/rfc/rfc2818.html#section-3.1) & [RFC6066](https://datatracker.ietf.org/doc/html/rfc6066#section-3)

wangw-1991:
Hi, @mfoltzgoogle. Can you help review this? Thanks.

backkem:
TODO:
* [ ] Update to match resolution for #275.
* [ ] Verify that TLS mandates the connection is closed if the SNI doesn't match.

markafoltz:
I updated the PR based on our discussion from TPAC with a slight adjustment.  I retained the counter in the agent serial number to make it easier to implement certificate rotation in the future (it won't require the agent to generate a new UUID).

PTAL @backkem

backkem:
Looks good to me. It addresses both the circularity and SNI issue. (I can't add a review, maybe because I opened the PR.)

Thanks for picking this up Mark. Sorry I didn't find the time yet.

markafoltz:
Thanks.  I am going to revert the SVG changes and submit them separately, as the SVG needs to be updated to reflect the updated PR.

Issue #334, opened 2024-05-20
Add guidelines on how CBOR messages are mapped onto streams


markafoltz:
Provisionally, each CBOR message gets mapped onto one QUIC stream.  This allows the message to be delivered to the agent as soon as it is fully received.  However, this is not made explicit in the spec.

In QUIC, there are per-connection limits on the number of simultaneous streams.   We may need to group multiple messages into a single stream if the limits are too low.  Need to research what the practical limits are in current implementation (i.e. QUICHE).

https://www.rfc-editor.org/rfc/rfc9000.html#section-4

mfoltzgoogle:
MAX_STREAMS is negotiated between peers as a transport parameter.  We should just tell agents to set MAX_STREAMS to a limit that is high enough so that it does not limit the protocol in practice.

An application sending 1 message/ms and allowing retransmissions for 10 seconds would have 10,000 concurrent streams.  We can just set a minimum MAX_STREAMS of 2^17 to provide additional headroom.

mfoltzgoogle:
Correction:   MAX_STREAMS limits the *cumulative* number of streams that can be opened per connection.   For long lived connections, we will need a higher value.  For a year-long connection, that could be as high as 400M according to the math above.

wangw-1991:
In current Open Screen Library, QuicSteam is used as unidirectional stream, see [CreateOutgoingStream](https://source.chromium.org/chromium/chromium/src/+/main:third_party/openscreen/src/osp/impl/quic/open_screen_session_base.cc;l=63) and [CreateIncomingStream](https://source.chromium.org/chromium/chromium/src/+/main:third_party/openscreen/src/osp/impl/quic/open_screen_session_base.cc;l=84;). One endpoint can only use the incoming stream to read message. If it want to send message to peer, it need to create a new outgoing stream.

If we can change the QuicStream to bidirectional, then the incoming stream can also be used to send message and there is no need to create a new one. Beside, we can also call [CloseWriteSide](https://source.chromium.org/chromium/chromium/src/+/main:net/third_party/quiche/src/quiche/quic/core/quic_stream.h;l=472;bpv=0;bpt=1) or [CloseReadSide](https://source.chromium.org/chromium/chromium/src/+/main:net/third_party/quiche/src/quiche/quic/core/quic_stream.h;l=467;bpv=0;bpt=1) to turn a bidirectional stream to a unidirectional stream for scenario where unidirectional stream is really needed. WDYT?

[...]

PR #336, opened 2024-06-01
Add a NOTE about MAX_STREAMS for QUIC connections

markafoltz:
Addresses #334. 

MAX_STREAMS limits the number of QUIC streams that can be opened per connection.  Recommend a high value so that agents won't run into this limit in practice.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/openscreenprotocol/pull/336.html" title="Last updated on Sep 19, 2024, 6:44 PM UTC (880b286)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/openscreenprotocol/336/9f704d4...880b286.html" title="Last updated on Sep 19, 2024, 6:44 PM UTC (880b286)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -43,7 +43,6 @@ url: https://datatracker.ietf.org/doc/html/rfc9000#name-transport-parameter-enco
 url: https://datatracker.ietf.org/doc/html/rfc9000#name-connection_close-frames; type: dfn; spec: RFC9000; text: CONNECTION_CLOSE Frames
 url: https://datatracker.ietf.org/doc/html/rfc9000#name-variable-length-integer-enc; type: dfn; spec: RFC9000; text: Variable-Length Integer Encoding
 url: https://datatracker.ietf.org/doc/html/rfc9000#name-variable-length-integer-enc; type: dfn; spec: RFC9000; text: variable-length integer
-url: https://datatracker.ietf.org/doc/html/rfc9000#section-4.6; type: dfn; spec: RFC9000; text: max_streams
 url: https://tools.ietf.org/html/rfc6762#section-9; type: dfn; spec: RFC6762; text: conflict resolution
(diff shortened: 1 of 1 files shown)


mfoltzgoogle:
@DavidSchinazi am I interpreting the QUIC RFC correctly here?  Appreciate it if you could take a look.

DavidSchinazi:
MAX_STREAMS frames exist for flow control purposes. This isn't something that is set once per connection; an endpoint will repeatedly send MAX_FRAMES with increasing values to its peer in order to keep the data flowing.

Here's an example: let's say I have limited memory and only want to allow 100 simultaneous streams. I'll start off by sending you MAX_STREAMS with 100, you'll open up some streams, and let's say when you have closed your first 10 streams, I'll send you another MAX_STREAMS, but this time with 110 - that way you always have 100 possible. Many other strategies are possible. You could for example start off with a small limit, and then increase if you deem the peer trustworthy.

But more fundamentally, this is not something application protocols should worry about. The statement "configure your QUIC stream limits such that it doesn't get in the way of your application performance" is true for all application protocols, there's no point in having every application protocol repeat it. Because once you go down that route, it's pretty much endless. You could also say "make sure your delayed ack strategy doesn't harm performance", and similar guidance for many other internal details of QUIC too.

What you do need to specify is whether you use unidirectional vs bidirectional streams, but looks like you covered that in 4581b3da4661b90636e9c5cec09f710a60f50182.

markafoltz:
Thanks for the feedback @DavidSchinazi.  I've updated the NOTE with simpler advice for implementations along the lines of what you suggested.

DavidSchinazi (approved):

markafoltz:
Per https://www.w3.org/2024/09/27-webscreens-minutes.html#t03:

> RESOLUTION: Close issue #334 by merging PR #336

Issue #341, opened 2024-09-16
[Meta] Open Screen Application Protocol


markafoltz:
Following TPAC 2023, we agreed to investigate splitting the current specification into two or more specifications, for several reasons.  Three main factors were:

- Simplifying the specification and making it easier to digest;
- Permitting multiple implementations at the transport layer to accommodate new device interconnection strata like WebTransport and Matter;
- Allowing application level concerns to be worked on in the W3C, while security sensitive networking  and certificate management could be handled in other fora in the future, like the IETF.

This meta issue tracks work items to deliver the application layer protocol, which is layered above DNS-SD, QUIC, and TLS.  Meaning, it can be implemented with an alternative to those technologies, if that alternative meets the necessary requirements.

backkem:
Side-note: I'm happy to accommodate for rapid experimentation in [go-lp2p](https://github.com/backkem/go-lp2p) if it makes sense at any point.

markafoltz:
The intention of the Open Screen Application Protocol is to work across multiple discovery, authentication, and transport mechanisms.    As we craft requirements for those mechanisms, I look forward to your feedback as to whether the various alternatives you're investigating would be suitable.  This work is tracked in Issue #342. 

I don't see anything to "argue" here.

markafoltz:
One more detail.  It's an open question whether authentication and certificate creation and exchange can be specified by the application protocol, or whether it is tied too closely to the networking aspects to make that possible.  That is an area that deserves a deeper dive.

In general "local https" approaches were not considered viable at the time we started this work, because it was quite complicated to get a CA to issue a certificate that essentially names a local network device that browsers would trust.  However there's likely been progress in this area since we started this work.

markafoltz:
The draft above does not include the authentication protocol (SPAKE2), but it could be added back depending on further discussions.

backkem:
I saw Mark hinted at a JavaScript implementation of the application protocol in the TPAC 2024 slides and wanted entertain a hypothetical full-stack version of the concept:
* Use the [Local Peer-to-Peer API](https://github.com/WICG/local-peer-to-peer) for peer discovery & authentication. Allowing the user agent to open secured connections between the peers.
  * Using the "OSP network protocol", Matter or others.
* Use this to open a [RTCQuicTransport](https://w3c.github.io/p2p-webtransport/#rtcquictransport*) to transport control messages.
[...]

PR #344, opened 2024-09-18
Initial draft of the Open Screen Application Protocol.

markafoltz:
Initial draft of the application side of the spec, following plan in #341.


backkem (approved):
This is a clean initial split.

It may be worth considering if part of the `agent-info` concerns are better handled on the 'network' side of the split. E.g.: it's hard to know what you're connecting to/authenticating with without giving it a proper name. But this can be refined in iterations.

application.bs
```diff
@@ -0,0 +1,2350 @@
+<pre class='metadata'>
+Title: Open Screen Application Protocol
+Shortname: openscreenprotocol-application
+Level: None
+Status: w3c/ED
+ED: https://w3c.github.io/openscreenprotocol/application.html
+TR: https://www.w3.org/TR/openscreenprotocol/application.html
```
tidoust, line 7: We won't immediately publish the spec to /TR. And when we do, it will use the `openscreenprotocol-application` shortname, and thus be published at a different URL.

I think the line can be dropped for now.

```suggestion
```

application.bs
```diff
@@ -0,0 +1,2350 @@
...
+
+The Open Screen Application Protocol is intended to be supported by a variety of
+connection technologies between browsers and devices.  There are specific
+requirements to allow browsers and devices to discover and authenticate to one
+another, but these requirements may be met by multiple possible implemenations.
+To maximize interoperability, browsers and devices should support the Open
+Screen Network Protocol, which provides a way for browsers and devices to
+discover, connect, and authenticate each other on a local area network.
+
+It allows a browser to present a URL, initiate remote playback of an HTML <l
```
tidoust, line 65: What is "it" here? The previous paragraph mentions the Open Screen Network Protocol, but I suspect this is more the Open Screen Application Protocol.
```suggestion
The Open Screen Application Protocol allows a browser to present a URL, initiate remote playback of an HTML <l
```

application.bs
```diff
@@ -0,0 +1,2350 @@
...
+spec=html>[=media element=]</l>, and stream media data to another device.
+
+The Open Screen Protocol is intended to be extensible, so that additional
```
tidoust, line 68: Similarly, the "Open Screen Protocol" would no longer exist (or would need to be defined as the combination of the Open Screen Application Protocol and the Open Screen Network Protocol).
```suggestion
The Open Screen Application Protocol is intended to be extensible, so that additional
```

tidoust (changes requested):
I just spotted a couple of metadata updates to do.
I haven't reviewed the entire `application.bs` file but that seems a good start :)

application.bs
```diff
@@ -0,0 +1,2350 @@
...
+============================
+
+The Open Screen Application Protocol connects browsers to devices capable of
+rendering Web content for a shared audience.  Typically, these are devices like
+Internet-connected TVs, HDMI dongles, and smart speakers.
+
+The Open Screen Application Protocol is intended to be supported by a variety of
+connection technologies between browsers and devices.  There are specific
+requirements to allow browsers and devices to discover and authenticate to one
+another, but these requirements may be met by multiple possible implemenations.
```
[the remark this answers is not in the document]
markafoltz: Done

markafoltz:
> It may be worth considering if part of the agent-info concerns are better handled on the 'network' side of the split. E.g.: it's hard to know what you're connecting to/authenticating with without giving it a proper name. But this can be refined in iterations.

@backkem I noted this while working on the network side of the spec and filed #346, will post your comment there.

application.html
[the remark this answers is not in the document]
markafoltz: Done.   I also added it to .gitignore

application.bs
markafoltz: Done

application.bs
markafoltz: Done

tidoust (approved):

anssiko (approved):
Thank you @markafoltz for getting the split drafts out ahead the meeting.

I'd be supportive of merging this and #347 with an Editor's Draft status to facilitate review. The group decision time is when we want to publish to TR. The local builds are clean, but it'd be convenient to be able to hand out URLs to rendered documents to reviewers outside the group as well, and deploy would enable that.

@tidoust build/validate/deploy for multiple specs since spec-prod [v2.3.0](https://github.com/w3c/spec-prod/releases/tag/v2.3.0) do work and the workflows look good to me, thanks for tweaking those. As a secondary non-blocking consideration: PR Preview does not yet support repos with multiple specs https://github.com/tobie/pr-preview/issues/18 -- are you aware of workarounds (besides separate repos)? Maybe we can live without PR Preview for a while...

anssiko:
@tidoust `application.html` deploy to gh-pages was [skipped](https://github.com/w3c/openscreenprotocol/actions/runs/10964676411/job/30448807503?pr=344#step:7:969) resulting in https://w3c.github.io/openscreenprotocol/application.html

Build and validate passed.

Looking at https://github.com/w3c/spec-prod/blob/main/docs/examples.md#multiple-specs-in-same-repository I believe a possible fix is to add this line:
`DESTINATION: application.html`

After this line: https://github.com/w3c/openscreenprotocol/blob/659582c1c0ae9e1775440556ab9ade008cae3a89/.github/workflows/auto-publish.yml#L57

(And similarly for the `network` part.)

tidoust:
> I believe a possible fix is to add this line: `DESTINATION: application.html`

That may work but, if it does, I don't understand why ;)
[`DESTINATION`](https://w3c.github.io/spec-prod/#destination) should take the value `application.html` all by itself when `SOURCE` is `application.bs`.

I'm looking into it but don't know yet why the job skipped deployment to gh-pages.

tidoust:
> I'm looking into it but don't know yet why the job skipped deployment to gh-pages.

Actually, that's easy, we're just not looking at the right logs ;) These logs are for the PR itself, where the spec-prod action validates things but indeed does not deploy anything (on purpose, that's a PR). The right logs to look show a [Bikeshed validation error](https://github.com/w3c/openscreenprotocol/actions/runs/10964712449/job/30448925870) because nothing build `application_message.html`. That's what needs fixing. I'll prepare a PR.

tidoust:
> The right logs to look show a [Bikeshed validation error](https://github.com/w3c/openscreenprotocol/actions/runs/10964712449/job/30448925870) because nothing build application_messages.html

More accurately, the file gets generated, but the action resets the status of the git folder after taking care of the first document (through a call to `git clean`), which deletes the other generated HTML files and makes the following generation of `application.html` fail. The action works best in completely separate jobs. Anyway, I'll propose a rewrite.

tidoust:
Publication now fixed. Both drafts are now available at:
https://w3c.github.io/openscreenprotocol/application.html
https://w3c.github.io/openscreenprotocol/network.html

PR #348, opened 2024-09-23
Fix generation of application and network specs

tidoust:
The w3c/spec-prod action cleans files at the end of a build, and drops the CDDL messages appendices files that need to be imported to build the application and network specs.

Workaround here is to re-generate the messages files each time.

Fix for https://github.com/w3c/openscreenprotocol/pull/344#issuecomment-2364521197


tidoust:
Taking the liberty to merge so that new draft documents get (hopefully!) published. We can always refine later on as needed.

PR #353, opened 2024-11-29
Bump version of Ubuntu and actions

tidoust:
Bikeshed now requires Python 3.9+, which requires running jobs using a more recent version of Ubuntu.

### DIFF
--- a/.github/workflows/auto-publish.yml
+++ b/.github/workflows/auto-publish.yml
@@ -19,15 +19,15 @@ on:
 
 jobs:
   main:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
-      # See doc at https://github.com/actions/checkout#checkout-v2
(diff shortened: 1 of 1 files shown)


anssiko (approved):
I see this is about Bikeshed workflow deps, but still want to ask:

Do the ReSpec workflows work with the latest or depend on the older ubuntu-20.04 and checkout@v2?

(We're setting up some ReSpec workflows elsewhere and this PR conveniently came up so wanted to ask in this context.)

tidoust:
> Do the ReSpec workflows work with the latest or depend on the older ubuntu-20.04 and checkout@v2?

I may be wrong but I don't think that there was ever a time when these versions were required. They just happened to be the latest versions at the time things got written. In any case, all examples in the spec-prod [repo](https://github.com/w3c/spec-prod) and [documentation](https://w3c.github.io/spec-prod/) now use `ubuntu-latest` and `checkout@v4`.

PR #356, opened 2024-12-11
Add discontinued note, link to /TR, update refs

tidoust:
This adds a note to the Status of This Document section of the previous Open Screen Protocol to note that the work was discontinued and redirect to the split documents (I copied the entire section because that makes life easier for Bikeshed, we shouldn't need to update that document any time soon in any case).

This also updates the references to IETF documents used in the Application and Network documents, and adds the links to the /TR versions.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/openscreenprotocol/pull/356.html" title="Last updated on Dec 11, 2024, 2:37 PM UTC (a78a59d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/openscreenprotocol/356/ed2685d...a78a59d.html" title="Last updated on Dec 11, 2024, 2:37 PM UTC (a78a59d)">Diff</a>

PR #357, opened 2024-12-11
Add missing parameters to enable auto-publication of split docs

tidoust:
The spec-prod action needs a couple of additional parameters that weren't set.

I'll add the Echidna tokens once #356 is merged (needed to have the right parameters in the documents themselves)


anssiko (approved):

tidoust:
I added publication tokens as secrets. Next update to the specs should also update the /TR versions. If that fails, well, something's wrong ;)

PR #361, opened 2025-06-09
Avoid naming confusion.

backkem:
Small rename to avoid confusion.


anssiko (approved):
Thanks! The build errors seem unrelated to this change.

@tidoust, this is the first change after the FPWD snapshot. Do we need to adjust the boilerplate to fix CI/CD?

backkem:
Build fixed in #362.

anssiko:
OK, it seems the build issue was just some markup typos, now fixed in #362. Rebase should turn this green.

anssiko:
(This var is internal to the agent hostname algorithm, thus the rename has no side effects, consider safe to merge.)

